PR #1, opened 2026-04-20
Add basic shape for Lifecycle Manager repository

ipetrov117:
This PR introduces the basic shape of the Lifecycle Manager repository, it introduces the following changes:

- Basic Kubernetes controller skeleton ([4c07dc6](https://github.com/SUSE/elemental-lifecycle-manager/commit/4c07dc6155bddeb9022519fbd854d704cd2c16bc), [7cb2e5b](https://github.com/SUSE/elemental-lifecycle-manager/commit/7cb2e5baf4a9272a77bdf4310b8341ad9dd21c4f)) - Generated by the Kubebuilder [init](https://book.kubebuilder.io/quick-start.html#create-a-project), [create api](https://book.kubebuilder.io/quick-start.html#create-an-api) and [create webhook](https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation.html#implementing-defaultingvalidating-webhooks) commands. The output of these commands was reduced to only include viable to this project configurations (e.g. Prometheus integration has been removed for the moment). In addition to this unit test and linter workflows were pinned to specific versions for better security compliance.
- Repo shape improvements:
  - Enable dependabot ([d6a6619](https://github.com/SUSE/elemental-lifecycle-manager/commit/d6a66193796bf537fbe9ef33b16930af81874a02))
  - Add repository license ([db1c076](https://github.com/SUSE/elemental-lifecycle-manager/commit/db1c076e245b5167f81d092a8b6a172e79cd8d27)) - adds the default license as well as linter license hardening 
  - Add basic README.md ([749631c](https://github.com/SUSE/elemental-lifecycle-manager/commit/749631cfa3fa41a8c09af4ea0d4eb112c57ebbe2)) - a more complex README will be created once the main Lifecycle Manager logic lands.
  - Add repo CODEOWNERS ([ab795f4](https://github.com/SUSE/elemental-lifecycle-manager/commit/ab795f4a721950032034c74ce65fa70868a5ce82))
- Actual new Lifecycle Manager logic introduced:
  - Updated the Release resource API ([11e69b3](https://github.com/SUSE/elemental-lifecycle-manager/commit/11e69b315cad527c6d51eb0b6c0adb395f663481), [e2525ba](https://github.com/SUSE/elemental-lifecycle-manager/commit/e2525ba8f989a008cbdb220cfa34d22d58d6d8c0)) - The API has been updated to include the main configurations and statuses that the custom resource will hold. Furthermore, the main condition statuses and reasons have also been introduced.
  - Add the necessary permissions so that the Lifecycle Manager can operate on the cluster ([985b3f5](https://github.com/SUSE/elemental-lifecycle-manager/commit/985b3f553ea889e2f7c069ab95f60babb6f14010)) - These are all the permissions that the Lifecycle Manager should need once the full logic lands. It was opted to include them here in one batch in order to not accidentally miss something throughout the upcoming PRs.
  - Release resource validation webhook implementation ([25eb180](https://github.com/SUSE/elemental-lifecycle-manager/commit/25eb18076a2bd0f039fbebe7393eb1ba7f5785a8)) - The validation webhook mainly ensures that there is only one Release resource in the cluster and that this resource can only be updated if there is not an already running upgrade process)
  - Skeleton for the reconciliation loop of the controller ([be1e1a6](https://github.com/SUSE/elemental-lifecycle-manager/commit/be1e1a61c396b7d238574f4f500c625d22ea4895)) - provides a working bare bones reconciliation loop that ensures that a Release resource is only updated if a change to the resource has actually occurred. This ensures that we refer to the latest Release resource state, effectively mitigating much of the errors regarding not updating the latest resource state.

What will follow after this PR?
- Release Manifest Parsing logic
- General upgrade implementation structure
- OS reconciliation logic
- K8s reconciliation logic
- Helm reconciliation logic

What will be handled once the initial Lifecycle Manger logic has landed (an Issue will be created for each point in due time)?
- Comprehensive unit tests
- Comprehensive README.md update
- Pin of mutable Makefile dependencies (lint, kustomize, controller-tools, etc.)
- Functionality improvements


internal/webhook/v1alpha1/release_webhook.go
```diff
@@ -32,50 +37,115 @@ import (
 var releaselog = logf.Log.WithName("release-resource")
 
 // SetupReleaseWebhookWithManager registers the webhook for Release in the manager.
```
dharmit:
```suggestion
// SetupWebhookWithManager registers the webhook for Release in the manager.
```

internal/webhook/v1alpha1/release_webhook.go
```diff
@@ -32,50 +37,115 @@ import (
...
+// as a generic version format. An error is returned if both attempts fail.
+func validateReleaseVersion(releaseVersion string) (*version.Version, error) {
+	if releaseVersion == "" {
+		return nil, fmt.Errorf("release version is required")
+	}
 
-	// TODO(user): fill in your validation logic upon object deletion.
+	v, err := version.Parse(releaseVersion)
+	if err != nil {
+		return nil, fmt.Errorf("'%s' is not a semantic version", releaseVersion)
```
dharmit:
```suggestion
		return nil, fmt.Errorf("%q is not a semantic version", releaseVersion)
```

dharmit (approved):
Couple of good-to-haves.

.github/dependabot.yml
```diff
@@ -0,0 +1,15 @@
+---
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    labels:
+      - "kind/dependabot"
```
frelon, line 7: nit: Label does not exist in the repo

go.mod
```diff
@@ -0,0 +1,100 @@
+module github.com/suse/elemental-lifecycle-manager
+
+go 1.25.6
```
frelon, line 3: Should we upgrade to 1.26?

frelon (approved):

.github/dependabot.yml
ipetrov117: Great catch! I added the label in the repo.

go.mod
ipetrov117: We probably should. Will update the version and do a quick test to be sure nothing breaks.

dharmit: While I don't think it's technically required, does it make sense to keep the go versions in sync between this and the elemental repos? It's not a rhetorical question — I'm just curious/wondering.

ipetrov117: Since LCM is(will be) effectively a consumer of the elemental library, I'd say it is fine for it to use a newer version. A theoretical problem can occur if it is specifying an older version. Then we could hit all sort of potential toolchain problems. That said, apart from a convenience standpoint, I do not think that there is a technical requirement for the versions to match. I'd prefer using the newer version as it will contain the most up-to date toolchain bug/security fixes, etc.

api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -0,0 +1,123 @@
+//go:build !ignore_autogenerated
```
dirkmueller, line 1: is this supposed to be in git? what is generating it?

Dockerfile
```diff
@@ -0,0 +1,31 @@
+# Build the manager binary
+FROM golang:1.25 AS builder
```
dirkmueller: please use registry.suse.com/bci/golang:1.25 instead.

Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# Build
+# the GOARCH has no default value to allow the binary to be built according to the host where the command
+# was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
+# the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
+# by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o manager cmd/main.go
+
+# Use distroless as minimal base image to package the manager binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/static:nonroot
```
dirkmueller: could be registry.suse.com/bci/bci-micro instead, or from scratch (but then requires a static build)

go.mod
dirkmueller: both are currently maintained, so they should be equal.

api/v1alpha1/zz_generated.deepcopy.go
ipetrov117: Not sure to be honest, but this is a file created and managed strictly by `controller-gen` (which we call through [make generate](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/initial-controller/Makefile#L48-L50)). We should not make any edits over this file as stated a couple of lines later in the file:
```
// Code generated by controller-gen. DO NOT EDIT.
```

Edit: Ah, I just realized you meant the whole file, not that specific line. Yes, the file is generated, but it is also part of the Go package needed to compile the API types. If it is not committed, developers and CI need to run `controller-gen` before the project can build. It is a common practice to commit this file, as it keeps the repo buildable from a fresh clone and makes development easier.

Dockerfile
ipetrov117: Yep, that was pre-built from kubebuilder, will set it up for the suse specific golang version.

Dockerfile
ipetrov117: The build is already static as it is being run with the `CGO_ENABLED=0` variable. I'll use `scratch`. Thanks!

go.mod
ipetrov117: Yea that is a good point. Let's keep the 1.25.6 version then, bumping will just cause confusion at this point. We can bump the version later down the line.

Dockerfile
ipetrov117: Image base updated to use `scratch` instead of `gcr.io/distroless/static:nonroot`.

Dockerfile
ipetrov117: Changed to use `registry.suse.com/bci/golang:1.25`. Thanks for the feedback!

api/v1alpha1/zz_generated.deepcopy.go
dirkmueller: fine, thanks for the explanation. we need to have a ci/cd eventually for this, but not blocking.

dirkmueller (approved):

PR #3, opened 2026-04-21
Introduce release manifest resolution logic

ipetrov117:
This PR introduces the logic that the Lifecycle Manager will use for release manifest resolution.

The manifest is resolved by using the [elemental resolution logic](https://github.com/SUSE/elemental/tree/main/pkg/manifest/resolver). The resolved manifest is stored by version in a ConfigMap cache. This cache is then called for each reconciliation and if the manifest is already present, it is retrieved from that cache and is not pulled again.

Additional unit tests for this logic will be implemented once the full LCM logic lands.

- General upgrade implementation structure + Release resource conditions handling
- OS reconciliation logic
- K8s reconciliation logic
- Helm reconciliation logic


internal/release/manifest_cache.go
```diff
@@ -0,0 +1,105 @@
...
+	if err := json.Unmarshal([]byte(data), manifest); err != nil {
+		return nil, fmt.Errorf("unmarshaling cached manifest: %w", err)
+	}
+
+	return manifest, nil
+}
+
+// Set stores a manifest in the cache for the given release version.
+func (c *ManifestCache) Set(ctx context.Context, namespace, version string, manifest *resolver.ResolvedManifest) error {
+	data, err := json.Marshal(manifest)
```
dharmit, line 72: Would a `nil` check for `manifest` add value here? Maybe there's no way `Set` would be call that way and we might not need it.

internal/controller/release_controller.go
```diff
@@ -31,6 +32,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 
 	lifecyclev1alpha1 "github.com/suse/elemental-lifecycle-manager/api/v1alpha1"
+	manifestCache "github.com/suse/elemental-lifecycle-manager/internal/release"
```
dharmit: I understand this is done because we're using the `ManifestCache` from the underlying package, but it kinda feels wrong to call it `manifestCache`.

dharmit (approved):
Only a couple of comments. LGTM otherwise.

internal/release/manifest_cache.go
ipetrov117: Since the manifest cache is intended to be called only from the reconciler and the reconciler will ensure that the manifest is [retrieved](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/release_manifest_resolution/internal/controller/release_controller.go#L152C1-L156) or fail otherwise, I'd say that for the moment we can skip the nil check here and assume that the caller will ensure that the manifest exists. If we get to a point where the manifest cache is called from a different place we can revisit. I did not want to over engineer this when I wrote it.

Does this make sense to you?

internal/controller/release_controller.go
ipetrov117: Fair, what name would be a better fit in your opinion? Perhaps `releasecache`? 

P.S.: I also wrongly added it as camelCase, where it should be lowercase only. So good catch.

internal/release/manifest_cache.go
dharmit: Absolutely! Thanks for explaining. 🙏🏽

internal/controller/release_controller.go
dharmit: > Perhaps `releasecache`?

Just `release` also works fine. But +1 for `releasecache` too.



> I also wrongly added it as camelCase

Ah, it's only the file names that must be camelCase, package import names need not be, is it so?

> So good catch.

By fluke. 😂

ipetrov117: > Just release also works fine. But +1 for releasecache too.

`release` won't be possible, as the parameter in the [function](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/release_manifest_resolution/internal/controller/release_controller.go#L140) that calls this library is called `release`. Too many "releases", I'll go with `releasecache` 👍

ipetrov117: Alias renamed to `releasecache`. Thanks!

internal/release/manifest.go
```diff
@@ -0,0 +1,97 @@
...
+	imageReadCloser := mutate.Extract(img)
+	defer func() {
+		_ = imageReadCloser.Close()
+	}()
+
+	tarReader := tar.NewReader(imageReadCloser)
+	for {
+		header, err := tarReader.Next()
+		if err != nil {
+			if err == io.EOF {
```
frelon:
```suggestion
			if  errors.Is(err, io.EOF) {
```

internal/release/manifest_cache.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	configMap := &corev1.ConfigMap{}
+	err = c.Client.Get(ctx, types.NamespacedName{
+		Name:      manifestCacheConfigMapName,
+		Namespace: namespace,
+	}, configMap)
+
+	if apierrors.IsNotFound(err) {
+		configMap = &corev1.ConfigMap{
```
frelon, line 84: Should we add an owner reference to the ConfigMap?

frelon (approved):

internal/release/manifest_cache.go
ipetrov117: The main benefit of having it would be to track resources during deletion of LCM and I thought about adding it, but ultimately decided to not overcomplicate the initial implementation. Once the initial implementation lands, I'll circle back to such "leftover" resources and figure out the best approach to cleaning them up (e.g. an owner reference, finalisers or something else).

That said, this is a valid comment and I have it under the ever-growing LCM "TODO" list.

PR #6, opened 2026-04-22
Introduce the basic upgrade reconciliation and condition logic

ipetrov117:
This PR introduces handling for future upgrade handler logic as well as condition updates for the Release resource.

Concepts introduced:
- [upgrade.Phase](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/upgrade_structure_and_conditions/internal/upgrade/phase.go#L28) - This is intended to depict the actual upgrade phase that the pipeline will be going through (e.g. OS, K8s, Helm, etc.).
- [upgrade.Pipeline](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/upgrade_structure_and_conditions/internal/upgrade/pipeline.go) - Introduces `PhaseHandlers` that  are responsible for different upgrade phases. Executes a set of handlers in a specific sequence and reports their status back to the controller.
- [upgrade.Config](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/upgrade_structure_and_conditions/internal/upgrade/config.go#L30) - This is the main instruction set that the pipeline and  future phase handlers will use to understand what changes need to be applied. Is constructed from multiple data points including: parsed release manifest, Release resource, controller specific logic.

Release resource API changes:
- [`spec.disableDrain`](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/upgrade_structure_and_conditions/api/v1alpha1/release_types.go#L68C27-L68C39) - Users can use this property to alter the default controller behaviour when it comes to node drains. By default the controller will look at the cluster nodes and automatically determine whether a drain for a specific node type can be performed (e.g. control-plane, worker, etc.). It then propagates this information to the `upgrade.Config` so that it can be transferred to future phase handlers which will actually schedule the drain itself. By specifying this property, users explicitly instruct the controller to not do any control/worker drains.

> Note: This is the last Release resource API change for the initial LCM logic.

Introduced the following Release `.status.conditions` handling:
- All upgrade components (e.g. Manifest resolution, PhaseHandlers) have their condition status tracked and reported on each reconciliation loop. Effectively allowing for a reliable tracking of the upgrade process.
- When the Release resource is updated all conditions are re-initialised to their initial `Pending` statuses - done to ensure to false information is given during update. 

- OS reconciliation logic
- K8s reconciliation logic
- Helm reconciliation logic

### DIFF
--- a/internal/controller/condition.go
+++ b/internal/controller/condition.go
@@ -1,0 +1,136 @@
+/*
+Copyright © 2026 SUSE LLC
+SPDX-License-Identifier: Apache-2.0
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controller
+
+import (
+	"errors"
+	"fmt"
+
+	lifecyclev1alpha1 "github.com/suse/elemental-lifecycle-manager/api/v1alpha1"
+	"github.com/suse/elemental-lifecycle-manager/internal/upgrade"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func setCondition(release *lifecyclev1alpha1.Release, conditionType string, status metav1.ConditionStatus, reason, message string) {
+	apimeta.SetStatusCondition(&release.Status.Conditions, metav1.Condition{
+		Type:               conditionType,
+		Status:             status,
+		ObservedGeneration: release.Generation,
+		Reason:             reason,
+		Message:            message,
+	})
+}
+
+// initializePendingConditions sets pending conditions for both the manifest and upgrade phases.
+func initializePendingConditions(release *lifecyclev1alpha1.Release, phases []upgrade.Phase) {
+	existing := apimeta.FindStatusCondition(release.Status.Conditions, lifecyclev1alpha1.ConditionManifestResolved)
+	if existing == nil {
+		setCondition(release, lifecyclev1alpha1.ConditionManifestResolved, metav1.ConditionFalse,
+			lifecyclev1alpha1.UpgradePending, "Waiting for release manifest to be resolved")
+	}
+
+	for _, phase := range phases {
+		conditionType := phase.ConditionType()
+		existing := apimeta.FindStatusCondition(release.Status.Conditions, conditionType)
+		if existing == nil {
+			setCondition(release, conditionType, metav1.ConditionFalse,
+				lifecyclev1alpha1.UpgradePending, "Waiting for previous phases to complete")
+		}
+	}
+}
+
+// setPhaseConditionFromError sets a failed condition based on a PhaseError.
+func setPhaseConditionFromError(release *lifecyclev1alpha1.Release, err error) {
+	var phaseErr *upgrade.PhaseError
+	if errors.As(err, &phaseErr) {
+		setCondition(release, phaseErr.Phase.ConditionType(), metav1.ConditionFalse,
+			lifecyclev1alpha1.UpgradeFailed, phaseErr.Err.Error())
+	}
+}
+
+// updatePhaseConditions updates the conditions for all phases from the reconciliation result.
+func updatePhaseConditions(release *lifecyclev1alpha1.Release, result *upgrade.Result) {
+	for phase, state := range result.PhaseStates {
+		updatePhaseCondition(release, phase, state)
+	}
+}
+
+// updatePhaseCondition updates the condition for a phase based on its observed status.
+func updatePhaseCondition(release *lifecyclev1alpha1.Release, phase upgrade.Phase, state *upgrade.PhaseStatus) {
+	conditionType := phase.ConditionType()
+	conditionStatus := metav1.ConditionFalse
+
+	if state.State == lifecyclev1alpha1.UpgradeSucceeded || state.State == lifecyclev1alpha1.UpgradeSkipped {
+		// Only update if not already succeeded or skipped to preserve timestamp
+		existing := apimeta.FindStatusCondition(release.Status.Conditions, conditionType)
+		if existing != nil && existing.Status == metav1.ConditionTrue {
+			return
+		}
+		conditionStatus = metav1.ConditionTrue
+	}
+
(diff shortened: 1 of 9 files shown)


internal/upgrade/config.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/suse/elemental/v3/pkg/manifest/resolver"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// Config represents a complete upgrade specification for all phases.
+type Config struct {
+	// ReleaseNamespacedName is the name and namespace of the Release resource.
+	ReleaseNamespacedName types.NamespacedName
+	// Version is the target release version.
+	Version string
```
dharmit: Could this be named `ReleaseVersion` akin to `ReleaseNamespacedName` so that it's clearer?

internal/upgrade/phase.go
```diff
@@ -0,0 +1,81 @@
...
+	PhaseKubernetes = Phase(strings.TrimSuffix(lifecyclev1alpha1.ConditionKubernetesUpgraded, "Upgraded"))
+	PhaseHelmCharts = Phase(strings.TrimSuffix(lifecyclev1alpha1.ConditionHelmChartsUpgraded, "Upgraded"))
+)
+
+// ConditionType returns the condition type string for this phase.
+func (p Phase) ConditionType() string {
+	return string(p) + "Upgraded"
+}
+
+// SkippedStatus retruns a skipped status for the given phase.
```
dharmit:
```suggestion
// SkippedStatus returns a skipped status for the given phase.
```

internal/controller/release_controller.go
```diff
@@ -162,6 +183,47 @@ func (r *ReleaseReconciler) getOrRetrieveManifest(ctx context.Context, release *
...
+			workerCounter++
+		}
+	}
+
+	switch {
+	case controlPlaneCounter > 1 && workerCounter <= 1:
+		return &upgrade.DrainOpts{ControlPlane: true, Worker: false}, nil
+	case controlPlaneCounter == 1 && workerCounter > 1:
+		return &upgrade.DrainOpts{ControlPlane: false, Worker: true}, nil
+	case controlPlaneCounter <= 1 && workerCounter <= 1:
```
dharmit, line 211: What kind of a setup would this be where control plane and worker count is 0?

internal/controller/condition.go
```diff
@@ -0,0 +1,136 @@
...
+			setCondition(release, conditionType, metav1.ConditionFalse,
+				lifecyclev1alpha1.UpgradePending, "Waiting for previous phases to complete")
+		}
+	}
+}
+
+// setPhaseConditionFromError sets a failed condition based on a PhaseError.
+func setPhaseConditionFromError(release *lifecyclev1alpha1.Release, err error) {
+	var phaseErr *upgrade.PhaseError
+	if errors.As(err, &phaseErr) {
```
frelon, line 61: What happens here if the error is not a `PhaseError`? Is it dropped silently?

internal/upgrade/phase.go
frelon:
```suggestion
// SkippedStatus returns a skipped status for the given phase.
```

frelon (approved):

internal/upgrade/pipeline.go
```diff
@@ -0,0 +1,87 @@
...
+	// Reconcile performs the reconciliation for this phase and returns the status.
+	Reconcile(ctx context.Context, config *Config) (*PhaseStatus, error)
+}
+
+// Pipeline executes upgrade phases in sequence, stopping when a phase
+// is not yet complete. This allows the controller to resume from where
+// it left off on the next reconciliation.
+type Pipeline struct {
+	handlers []PhaseHandler
+}
```
dharmit, line 40: The sequence is going to be premeditated, right? In the sense, every time LCM is first going to upgrade X, then Y, and then Z. So could/should that be hardcoded?

Am I thinking ahead here? Because `upgrade.NewPipeline` is called without any arguments/handlers in `cmd/main.go` and maybe this is going to become clearer in a future PR?

dharmit (commented):
Couple of questions because I struggled to understand few things in this PR.

Couple of nits/good-to-haves.

internal/controller/condition.go
ipetrov117: So this function is a best effort attempt to set a condition from an error. If the error is a `PhaseError` we set the condition, if the error is a different type, we cannot reliably map it to a phase, hence we cannot reliably map it to a condition message.

That said the error itself is not dropped silently, but is instead logged by the controller itself [here](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/upgrade_structure_and_conditions/internal/controller/release_controller.go#L132). 

That way if we encounter an error that we know how to turn to a Phase condition message, then we update the condition and log the error, otherwise we just log the error without a condition update, as we do not have a reliable way to map the error to a phase.

We could iterate over this in the future, but generally does that make sense to you?

internal/controller/release_controller.go
ipetrov117: This check is more oriented towards combining the following assertions:
1. Is the cluster consisting of 1 control plane and 1 worker
2. Is the cluster a single node control-plane cluster without workers

I decided to keep `<=` for the control-plane check, as I was not sure whether I was missing some edge use-cases and decided to be safe than sorry, if that makes sense.

internal/upgrade/pipeline.go
ipetrov117: Yes, so the idea here is to always run the PhaseHandlers in a specific order. The order of our setup would be: `OS PhaseHandler -> K8s PhaseHandler -> Helm PhaseHandler`. I can see how this can be hard to foresee from this PR, but I couldn't find a better way to split the logic.

Nevertheless, the end result would look something like this in `main.go`:

```go
Pipeline: upgrade.NewPipeline(
			reconcilers.NewOSReconciler(...),
			reconcilers.NewKubernetesReconciler(...),
			reconcilers.NewHelmReconciler(...),
		),
```

Where the pipeline would always schedule the reconciliation in that order - first OS, then K8s, then Helm. This will ensure that our upgrade will always follow the same upgrade steps.

Let me know whether this answers your question.

internal/upgrade/config.go
ipetrov117: Fixed 👍

internal/upgrade/pipeline.go
dharmit: Yes, that's exactly what I imagined it would be. Thanks for explaining.

internal/controller/release_controller.go
dharmit: Yeah, in general it makes sense, but the `controlPlaneCounter <= 1` has still got me lost. Maybe I'm just being pedantic/paranoid about "what if this check get satisfied erroneously". Anyway, we can handle that if/when that happens instead of being too imaginative now. 🤷🏽‍♂️ 😂 

Thanks again for explaining.

dharmit (approved):

PR #8, opened 2026-04-23
Introduce OS upgrade implementation

ipetrov117:
This PR introduces the initial LCM OS upgrade implementation.

Main concepts introduced:
- Base SUC Plan creation and reconciliation ([3445b59](https://github.com/SUSE/elemental-lifecycle-manager/commit/3445b59491c9dc8a2517d3321105e1d8f9d63af9), [2003e0d](https://github.com/SUSE/elemental-lifecycle-manager/commit/2003e0d718d97d10f413f575ff58a3119da3f8c3)) - Introduces the base way of creating an SUC Plan and reconciling it to a `PhaseStatus`. This logic will be used for both OS as well as Kubernetes upgrades.
- OS SUC Plan creation and reconciliation ([e2cfe44](https://github.com/SUSE/elemental-lifecycle-manager/commit/e2cfe440fa0810719802ec7bbfb582602e59a373)) - Introduces the SUC Plans specific to the OS upgrade as well as the reconciliation logic that handles those plans. 

General notes on the OS upgrade workflow:
1. Each OS upgrade process starts with a creation of SUC Plans for specific node types (e.g. `control-plane` and `worker` nodes). 
   > Note: `control-plane` and `worker` SUC Plans are **not** deployed together. `control-plane` SUC Plan are **always** deployed first and only **if** the `control-plane` SUC Plan completes successfully is then the `worker` SUC Plan created. This ensures that we **always** upgrade the OS of all control-planes before moving to doing any kind of upgrades over the worker nodes.
1. Both `control-plane` and `worker` SUC Plans will be creating Pods whose main container runs with the `uc-base-os-kernel-default` image version that is specified in the target release manifest.
1.  Both `control-plane` and `worker` SUC Plans will be checking whether the OS needs to be updated by looking at the `/etc/elemental/deployment.yaml` file. If an upgrade is necessary, both plans will be using the `upgrader` script from the container image to do the actual update. For the moment the update manually pulls the image instead of retrieving it from containerd, due to a limitation on how `elemental3ctl`'s container image handling logic works.
1. Fore `Release` resource updates, all previous SUC Plans will be removed before creating the new SUC Plans for the new `Release` version.
1. As SUC supports dynamic SUC Plan re-execution (e.g. a new node is added to the cluster and a previously completed plan becomes active again), LCM constantly reconciles the SUC Plans for the target release. If a Plan becomes active again, SUC will update the status of the `Release` resource and reconcile the SUC Plan until it once again completes. 

- K8s reconciliation logic
- Helm reconciliation logic


internal/plan/os.go
```diff
@@ -0,0 +1,144 @@
...
+
+	var out bytes.Buffer
+	if err := tmpl.Execute(&out, data); err != nil {
+		return "", fmt.Errorf("applying template for OS upgrade script: %w", err)
+	}
+
+	return out.String(), nil
+}
+
+func parseImage(image string) (repo, tag string) {
```
frelon: This is very alike the `parseImageTag` method in `internal/upgrade/config.go`, could we try to merge the two?

Or even use a library, I believe there are lot's of image/tag parsing methods out there.

internal/upgrade/reconcilers/plan.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return nodes, nil
+}
+
+// parsePhaseStatusFromPlan parses the SUC Plan's status to a PhaseStatus.
+func parsePhaseStatusFromPlan(p *upgradecattlev1.Plan) *upgrade.PhaseStatus {
+	if len(p.Status.Applying) > 0 {
+		return &upgrade.PhaseStatus{
+			State:   lifecyclev1alpha1.UpgradeInProgress,
+			Message: fmt.Sprintf("Plan %s is curently applying on: %s", p.Name, p.Status.Applying),
```
frelon:
```suggestion
			Message: fmt.Sprintf("Plan %s is currently applying on: %s", p.Name, p.Status.Applying),
```

internal/upgrade/reconcilers/os.go
```diff
@@ -0,0 +1,138 @@
...
+	return &upgrade.PhaseStatus{
+		State:   lifecyclev1alpha1.UpgradeSucceeded,
+		Message: "All nodes upgraded successfully",
+	}, nil
+}
+
+// preparePlans determines which types of SUC Plans does the cluster need and returns a list of ordered SUC plans ready for reconciliation.
+// Control plane plans are always ordered before worker plans, ensuring that all control-plane nodes are upgraded before any worker upgrade
+// operation starts.
+func (r *OSReconciler) preparePlans(ctx context.Context, config *upgrade.Config) (plans []*upgradecattlev1.Plan, err error) {
```
frelon: We are not using the named returns, can the names be dropped?

internal/controller/release_controller.go
```diff
@@ -224,10 +232,64 @@ func (r *ReleaseReconciler) parseUpgradeConfig(ctx context.Context, manifest *re
...
+			lifecyclev1alpha1.ReleaseVersionLabel: lifecyclev1alpha1.SanitizeVersion(oldVersion),
+		},
+	); err != nil {
+		return fmt.Errorf("listing old plans: %w", err)
+	}
+
+	for _, p := range planList.Items {
+		logger.Info("Deleting old plan", "name", p.Name)
+		if err := r.Delete(ctx, &p); err != nil {
+			return fmt.Errorf("deleting plan %s: %w", p.Name, err)
```
frelon: This returns on the first error, should we try to delete all plans and collect all errors to return instead?

internal/controller/release_controller.go
```diff
@@ -124,7 +127,12 @@ func (r *ReleaseReconciler) reconcileNormal(ctx context.Context, release *lifecy
 		return ctrl.Result{}, fmt.Errorf("parsing upgrade config: %w", err)
 	}
 
-	// TODO: cleanup SUC Plans for previous release
+	// Clean up plans from previous versions before creating new ones
+	if release.Status.Version != "" && release.Status.Version != config.ReleaseVersion {
+		if err := r.cleanupOldVersionPlans(ctx, release.Name, release.Status.Version); err != nil {
+			logger.Error(err, "Failed to cleanup old version plans", "oldVersion", release.Status.Version)
```
frelon, line 133: Here we only log the error, but continue execution. Should this log be a warning, or should we return an error here?

(In my eyes logging an error means we should not continue execution, but feel free to disregard)

internal/controller/release_controller.go
```diff
@@ -224,10 +232,64 @@ func (r *ReleaseReconciler) parseUpgradeConfig(ctx context.Context, manifest *re
...
+// mapPlanToRelease maps SUC Plan events to Release reconcile requests.
+// Uses the release name label on the Plan to find the corresponding Release.
+func (r *ReleaseReconciler) mapPlanToRelease(ctx context.Context, obj client.Object) []ctrl.Request {
+	releaseName := obj.GetLabels()[lifecyclev1alpha1.ReleaseNameLabel]
+	if releaseName == "" {
+		return nil
+	}
+
+	releaseList := &lifecyclev1alpha1.ReleaseList{}
+	if err := r.List(ctx, releaseList); err != nil {
```
frelon, line 277: Should we replace this with a Get to not have to list all objects on every call?

internal/plan/os.go
ipetrov117: I did not want to introduce a new library for such a small operation. That said, I see the duplication, I'll move the logic in a separate package so both `internal/upgrade/config.go` and `internal/plan/os.go` can access it. 

In the future if we start doing more complex image parsing/work, we could look to move to a library, but as it stands I prefer to not introduce a new dependency library for something this small.

internal/controller/release_controller.go
ipetrov117: I was looking at it from the PoV that if a single SUC plan cannot be deleted, then we should return as soon as possible, but I do see the reasoning in trying to delete all SUC plans and then concatenating the errors.

I'll fix this, thanks!

internal/controller/release_controller.go
ipetrov117: You are absolutely correct, we should not attempt to reconcile (e.g. create new SUC Plans) if the old ones are not cleaned up properly. Fixing this asap.

internal/controller/release_controller.go
ipetrov117: Yep, Get would be more lightweight than List, although List will only ever hold one value. I'll look to set this up

internal/plan/base_test.go
```diff
@@ -0,0 +1,82 @@
...
+		Expect(plan.TypeMeta.Kind).To(Equal("Plan"))
+		Expect(plan.TypeMeta.APIVersion).To(Equal("upgrade.cattle.io/v1"))
+
+		Expect(plan.ObjectMeta.Name).To(Equal("plan-1"))
+		Expect(plan.ObjectMeta.Namespace).To(Equal("cattle-system"))
+		Expect(plan.ObjectMeta.Labels).To(BeEmpty())
+		Expect(plan.ObjectMeta.Annotations).To(BeEmpty())
+
+		Expect(plan.Spec.ServiceAccountName).To(Equal("system-upgrade-controller"))
+		Expect(plan.Spec.Drain).To(BeNil())
```
dharmit, line 41: Shouldn't `Drain` be `false`? Expecting it to be `nil` seems inaccurate.

internal/plan/base_test.go
```diff
@@ -0,0 +1,82 @@
...
+		Expect(plan.Spec.ServiceAccountName).To(Equal("system-upgrade-controller"))
+
+		Expect(plan.Spec.Drain).NotTo(BeNil())
+		Expect(ptr.Deref(plan.Spec.Drain.DeleteEmptydirData, false)).To(BeTrue())
+		Expect(ptr.Deref(plan.Spec.Drain.IgnoreDaemonSets, false)).To(BeTrue())
+		Expect(plan.Spec.Drain.Force).To(BeTrue())
+		Expect(plan.Spec.Drain.Timeout.String()).To(Equal("15m"))
+	})
+
+	It("Properly creates a plan without draining options", func() {
```
dharmit: This test spec is already covered.

internal/upgrade/reconcilers/plan.go
```diff
@@ -0,0 +1,145 @@
...
+	plan, err := p.getOrCreatePlan(ctx, desired)
+	if err != nil {
+		return nil, fmt.Errorf("parsing plan: %w", err)
+	}
+
+	result := &PlanResult{Status: parsePhaseStatusFromPlan(plan)}
+	if result.Status.State == lifecyclev1alpha1.PlanComplete {
+		nodes, err := p.listNodesForPlan(ctx, plan)
+		if err != nil {
+			return nil, fmt.Errorf("listing plan '%s' nodes: %w", plan.Name, err)
```
dharmit:
```suggestion
			return nil, fmt.Errorf("listing plan %q nodes: %w", plan.Name, err)
```

internal/upgrade/reconcilers/plan.go
```diff
@@ -0,0 +1,145 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	return plan, nil
+}
+
+// listNodesForPlan lists the nodes that the specified SUC Plan is responsible for.
+func (p *SUCPlanReconciler) listNodesForPlan(ctx context.Context, plan *upgradecattlev1.Plan) (nodes *corev1.NodeList, err error) {
```
dharmit, line 101: Nothing wrong with the approach here, but wondering if instead of passing an entire `*plan`, would it be better and/or simpler to just pass the `plan.Spec.NodeSelector` that this method is actually using?

internal/controller/release_controller.go
ipetrov117: @frelon I remembered why I went with the List option. In order to do a `Get` instead of `List`, we would need to know the Release namespace, but currently that is not possible without setting up an additional label pointing to the namespace, because the Release can be deployed anywhere (e.g. the controller is not scoped to view a single namespace). That is why I chose to implement the logic with `List`, as it effectively mitigates the issue and keeps things fairly tight, mainly because there can only ever be one release object on the cluster (which is enforced by the webhook).

Nevertheless, I will be looking to scope LCM to watch a single namespace for Release objects. Would you consider an acceptable compromise to wait for the LCM namespace scope before switching to using the `Get` logic, that way we would not have to do any unnecessary additions of labels just to track a namespace. Sorry for the back and forth, too much code is flying by. Either way, please let me know.

internal/plan/os.go
ipetrov117: Moved the image parsing logic under the [`image`](https://github.com/ipetrov117/elemental-lifecycle-manager/tree/os_reconciliation_impl/internal/image) package and added some basic unit tests.

internal/controller/release_controller.go
ipetrov117: Fixed, we now attempt to delete all SUC plans before throwing any errors.

internal/upgrade/reconcilers/os.go
```diff
@@ -0,0 +1,138 @@
...
+	lifecyclev1alpha1 "github.com/suse/elemental-lifecycle-manager/api/v1alpha1"
+	"github.com/suse/elemental-lifecycle-manager/internal/plan"
+	"github.com/suse/elemental-lifecycle-manager/internal/upgrade"
+)
+
+// OSReconciler ensures that the operating system of the cluster nodes reflects the desired release state.
+type OSReconciler struct {
+	client.Client
+	sucReconciler PlanReconciler
+}
```
dharmit, line 38: Pardon my ignorance, but why should operating system reconciler contain information that's related to k8s? It might need it, but should it not be provided by something else?

internal/controller/release_controller.go
frelon: I see, as long as we leave a comment describing we only ever expert one Release object Im fine with the current implementation! 👍

frelon (approved):

internal/plan/os.go
```diff
@@ -0,0 +1,144 @@
...
+	return fmt.Sprintf("%s-%s", osControlPlaneBaseName, lifecyclev1alpha1.SanitizeVersion(version))
+}
+
+// osWorkerName returns the full plan name for the given version.
+func osWorkerName(version string) string {
+	return fmt.Sprintf("%s-%s", osWorkerBaseName, lifecyclev1alpha1.SanitizeVersion(version))
+}
+
+// OSControlPlane builds a SUC Plan for OS upgrades on control plane nodes.
+func OSControlPlane(releaseName, osImage, osVersion, releaseVersion string, drain bool) (*upgradecattlev1.Plan, error) {
```
dharmit, line 51: Lot of repetition happening between `OSControlPlane` and `OSWorker` functions. Even the signatures are same. Can we reduce it? Maybe an extra input parameter to differentiate between CP and worker nodes, and tweaks accordingly in implementation?

I sound to myself like seeking premature optimization. 😀

internal/plan/templates/os-upgrade.sh.tpl
```diff
@@ -0,0 +1,22 @@
+HOST="${HOST:-/host}"
```
dharmit, line 3: I think we should have these lines at the beginning:
```sh
#!/bin/bash

set -euo pipefail
```

internal/controller/release_controller.go
ipetrov117: Added a clarification comment. Thanks!

internal/upgrade/reconcilers/os.go
```diff
@@ -0,0 +1,138 @@
...
+func isControlPlaneOnlyCluster(nodes []corev1.Node) bool {
+	for _, node := range nodes {
+		if _, isControlPlane := node.Labels[plan.ControlPlaneLabel]; !isControlPlane {
+			return false
+		}
+	}
+	return true
+}
+
+func getNodeNamesFromList(nodes []corev1.Node) []string {
```
dharmit, line 131: Feel free to say it's a bad/wrong idea, but suggesting this to avoid making a copy. Not fighting for pointer usage, but it feels like we could benefit in a larger cluster, maybe. This might cascade all the way up to how `PlanResult` is defined. WDYT?
```suggestion
func getNodeNamesFromList(nodes *[]corev1.Node) []string {
```

internal/controller/release_controller.go
dharmit: It looks like original concern hasn't been addressed/fixed yet, or am I mis-reading things?

EDIT: I think I was looking at older commit.

dharmit (commented):
Mostly comments. But some might make sense for making changes so not "Approve"'ing yet.

internal/plan/base_test.go
ipetrov117: This is referring to the SUC's Plan [API](https://github.com/rancher/system-upgrade-controller/blob/v0.19.2/pkg/apis/upgrade.cattle.io/v1/types.go#L46), where `Drain` is a pointer to [DrainSpec](https://github.com/rancher/system-upgrade-controller/blob/v0.19.2/pkg/apis/upgrade.cattle.io/v1/types.go#L81), hence the nil check.

internal/plan/base_test.go
ipetrov117: This is a copy paste error, I added it to ensure the `Expect(plan.Spec.Cordon).To(BeTrue())` is set, I'll merge it with the top-level one.

internal/upgrade/reconcilers/plan.go
ipetrov117: I'd prefer to keep the full plan reference here to keep things easier to follow and extend in the future. Also since we refer to the pointer, we are not copying the plan to the function, but just referring it, so from a performance PoV, this should also be good.

dharmit: Sure, that's fine.

internal/upgrade/reconcilers/os.go
ipetrov117: It needs access to the K8s API, as it needs to be able to create the OS SUC Plan resources and reconcile their status. The SUC Plan resources are Kubernetes CRDs, that are deployed within the K8s cluster, hence the need for `client.Client`. 

Keep in mind that the actual logic that does the upgrade are the OS SUC Plans themselves, which are not tied to the Kubernetes API per-se, but are executed directly on each node. The OS reconciler is just the wrapper that deploys the SUC Plans and validates their status and in order for it to be able to do that it needs access to K8s.

internal/plan/os.go
ipetrov117: I did not want to merge them into a single function yet, as the control-plane SUC Plan could drastically differ from the worker SUC Plan. In the future, it might require additional tolerations, selectors and many more. I agree that they look very similar right now, but this could just as easily change fast, so I decided to not do any optimisations in this area right now. Let me know if that makes sense to you.

internal/plan/templates/os-upgrade.sh.tpl
ipetrov117: I do not think that this is necessary for the moment, mainly because:
1. There is no piping in the script (`-o pipefail`)
2. All variables have default values (`-u`)
3. And the only place that matters in terms of failures, which is the `upgrader`, actually [exits](https://src.suse.de/UnifiedCore/uc-factory/src/branch/main/OS/uc-base-os/upgrader.sh#L52) with the `elementalctl` status code.

So for the moment I'd like to keep the script as tight as possible, if that makes sense.

internal/upgrade/reconcilers/os.go
dharmit: That makes sense. Thanks for explaining.

internal/plan/os.go
dharmit: Yes, makes sense. Thanks for explaining.

internal/upgrade/reconcilers/os.go
ipetrov117: I don't think passing *[]corev1.Node buys us much here. A slice parameter only copies the slice header, not the backing array, so the nodes themselves are not copied when calling the function. 

If you want we can improve the underlying [for loop](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/os_reconciliation_impl/internal/upgrade/reconcilers/os.go#L133), right now it copies each `corev1.Node` value into `n` as it iterates, which can technically be wasteful when running on a larger cluster.

I can look to improve this to something like:

```go
for i := range nodes {
	nodeNames = append(nodeNames, nodes[i].Name)
}
```

That way we would not unnecessarily copy the node to `n` before using it, which would optimise the process for large-scale clusters. WDYT?

internal/plan/templates/os-upgrade.sh.tpl
dharmit: The `uo pipefail` was suggested to stay aligned with how scripts are written in elemental repo. But your arguments make sense.

I'd still request to add `-e` to sort of offload error handling to shell script itself lest we forget to add it in future or miss doing an error check.

A `#!/bin/bash` at the top is still required for a bash script, no?

dharmit: We had similar discussion for elemental repo: https://github.com/SUSE/elemental/pull/399#pullrequestreview-4020920927.

internal/upgrade/reconcilers/os.go
dharmit: > A slice parameter only copies the slice header, not the backing array, so the nodes themselves are not copied when calling the function.

You're right. 👍🏽 



> That way we would not unnecessarily copy the node to `n` before using it, which would optimise the process for large-scale clusters. WDYT?

This one didn't occur to me, but makes a lot of sense and doesn't even sound like premature optimization to me.

internal/plan/templates/os-upgrade.sh.tpl
ipetrov117: > A #!/bin/bash at the top is still required for a bash script, no?

This is not parsed to a script file, but is parsed to a [string](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/os_reconciliation_impl/internal/plan/os.go#L129) and passed to the SUC Plan's [`/bin/sh -c`](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/os_reconciliation_impl/internal/plan/os.go#L105) command. So, no for the current file we do not need to give the shebang.

> I'd still request to add -e to sort of offload error handling to shell script itself lest we forget to add it in future or miss doing an error check.

I would prefer to not do the `set -e` as that would mean that this [line](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/os_reconciliation_impl/internal/plan/templates/os-upgrade.sh.tpl#L6) will fail if there is no `uri: oci://$OS_IMAGE_REPO" "$DEPLOYMENT` string in the `/etc/elemenetal/deployment.yaml`. But for initial clusters this line will not return anything as there will be lie no such line, which will technically cause an error, but not have grep brake anything, just return an empty string. Which we validate [here](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/os_reconciliation_impl/internal/plan/templates/os-upgrade.sh.tpl#L11).

Our current setup is tricky, as for initial installations we do not install the OS from an OCI image but use RAW data. I tried to explain this in the comment [here](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/os_reconciliation_impl/internal/plan/templates/os-upgrade.sh.tpl#L8-L10). Furthermore we do not have many tools on the image to work with, so this is the best that I could do with the tools at hand. Let me know if this makes sense.

internal/plan/base_test.go
ipetrov117: Fixed 👍

internal/upgrade/reconcilers/os.go
ipetrov117: Cool, done 👍

internal/plan/templates/os-upgrade.sh.tpl
ipetrov117: > We had similar discussion for elemental repo: https://github.com/SUSE/elemental/pull/399#pullrequestreview-4020920927.

Sure, but right now nothing in the script fails unexpectedly or without being handled. And just adding `set -e` will break the existing script's logic as I explained in the above comment. I am not of the opinion that every script needs `set -e` as long as said script handles its errors.

ipetrov117: Added `set -e` and reworked the `os-upgrade.sh` file. 👍

dharmit: > Added `set -e` 

You didn't need to! Your explanations made sense. I was having my dinner so it took time to look at your helpful responses.

ipetrov117: Not necessary, I took a stab at the rework and it turns out it was a small change. So I am happy to keep it. Already tested and validated to be working. 

Also I think that you were right to point out the need for `set -e`, so let's keep it 🚀

dharmit: Marking resolved per Slack conversation. Thanks. 👍🏽

atanasdinov (approved):
Great! Thanks for all the efforts in getting this over the line.

internal/image/image.go
```diff
@@ -0,0 +1,43 @@
...
+*/
+
+package image
+
+import "strings"
+
+const defaultTag = "latest"
+
+// Parse splits an image reference into its repository and tag.
+func Parse(image string) (repo, tag string) {
```
frelon: FWIW we already have a dependency on `go-containerregistry` which is able to parse [image references](https://github.com/google/go-containerregistry/blob/d4f10504a3c9528aeb51c62c7a859cd0a47e07a8/pkg/name/ref.go#L40) and [tags](https://github.com/google/go-containerregistry/blob/d4f10504a3c9528aeb51c62c7a859cd0a47e07a8/pkg/name/tag.go#L77)

dharmit (approved):

internal/image/image.go
ipetrov117: Fantastic, I completely forgot we already had `go-containerregistry`, switched the image parsing logic to use it instead of the custom implementation. Thanks!

PR #10, opened 2026-04-24
Introduce Kubernetes upgrade implementation

ipetrov117:
> **Disclamer:** Parts of this logic will definitely change in future versions, as we align on the long-term Kubernetes upgrade process and move away from the short-term solution of using the [rke2-upgrade](https://github.com/rancher/rke2-upgrade) functionality.

This PR add Kubernetes distribution upgrade capabilities to the Lifecycle Manager.

**General upgrade workflow:**
The Kubernetes upgrade flow follows the same general pattern as the OS upgrade process, where SUC Plans are created for each node type and `control-plane` nodes are always upgraded before `worker` nodes.  The SUC Plans run the [rancher/rke2-upgrade](https://hub.docker.com/r/rancher/rke2-upgrade) images that ship a script that is capable of "upgrading" the Kubernetes distribution on a specific node. `rke2-upgrade` does the upgrade by replacing the existing `rke2` binary with a new binary for the desired version, killing the rke2 process and starting it again from the new binary.  As previously discussed, this follows the RKE2 approved process for [automated upgrades](https://docs.rke2.io/upgrades/automated).

**Validating core RKE2 packaged components**
During RKE2 upgrade, the controller also validates whether the core RKE2 packaged components (e.g. `rke2-coredns`, `rke2-ingress-nginx`, etc.)  are available before marking the Kubernetes upgrade as complete. 

_Why is this needed?_
Having such a validation in place ensures that the controller does not prematurely start a Helm chart upgrade while the Kubernetes core components are still becoming available. Without this in place, all sorts of issues can be hit. Mainly the Helm chart upgrade could fail for an increasingly long time, which on slower machines will cause a crash loopback that could last an increasingly long time, effectively delaying the upgrade process for an unforeseen amount of time. To mitigate this, the controller ensures the basic availability of the RKE2 packaged components before marking the Kubernetes upgrade as complete.

_Caveats_
Right now there is no conclusive way to get the list of components related to an RKE2 release from within the controller code. Because of this, the controller essentially tries to map the RKE2 components itself in the [rke2_core.go](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/kubernetes_reconciliation_impl/internal/upgrade/reconcilers/rke2_core.go) logic. This is a temporary fix and most of this logic will largely be removed when:
1. We come up with the long-term Kubernetes upgrade process
2. Come up with a more precise way to pinpoint the versions of a specific RKE2 K8s version (ideally we should communicate with the ECM team about this).

Nevertheless, for the short-term this implementation offers an acceptable compromise when it comes to mapping the RKE2 packaged components.

Screenshots:
_Before upgrade:_
Screenshot 2026-04-24 at 12 12 25

_During Kubernetes version upgrade:_
Screenshot 2026-04-24 at 12 13 10

_During core RKE2 packaged components availability validation:_
Screenshot 2026-04-24 at 12 14 50

_After upgrade:_
Screenshot 2026-04-24 at 12 16 28
Screenshot 2026-04-24 at 12 16 45

- Helm reconciliation logic
- Lifecycle Manager and CRD Helm charts


api/v1alpha1/release_types.go
```diff
@@ -67,6 +67,10 @@ const (
 
 	// PlanComplete indicates that a SUC Plan related to the upgrade process has completed.
 	PlanComplete = "PlanComplete"
+
+	// K8sPackagedComponentsAvailable indicates that all packaged components related to a
+	// Kubernetes distribution are available.
+	K8sPackagedComponentsAvailable = "Available"
```
atanasdinov: Is it a good idea to have a more descriptive constant for this state? IIUC, "Available" is usually the generic form for a final state of a successfully reconciled resource.

ipetrov117: It might be a good idea, yes, I kept it short, as it will only be internally checked and not shown directly in the conditions, but nevertheless we can have a more descriptive name.

How about `PackagedComponentsAvailable`?

atanasdinov: That looks perfect, thank you!

ipetrov117: Name changed. Thanks!

atanasdinov (approved):
Outstanding!

cmd/main.go
```diff
@@ -170,13 +175,43 @@ func main() {
...
+		os.Exit(1)
+	}
+
+	serverVersion, err := discoveryClient.ServerVersion()
+	if err != nil {
+		setupLog.Error(err, "unable to parse server version")
+		os.Exit(1)
+	}
+
+	var packagedComponendsHandler reconcilers.KubernetesPackagedComponentsHandler
```
frelon:
```suggestion
	var packagedComponentsHandler reconcilers.KubernetesPackagedComponentsHandler
```

internal/upgrade/reconcilers/rke2_core.go
```diff
@@ -0,0 +1,417 @@
...
+// RKE2 ships its core packaged components as HelmChart resources, so this function returns a map
+// of HelmChart resources keyed by the HelmChart name.
+func findRKE2PackagedComponents(ctx context.Context, c client.Client) (map[string]helmv1.HelmChart, error) {
+	// Right now there is no conclusive way to retrieve a RKE2 packaged component by a single delimeter, so this function
+	// uses a set of delimeters that are specific to RKE2 HelmChart resources. While it is possible to miss a some charts,
+	// this set of delimeters matches all HelmCharts related to the core use-case deploymet of RKE2.
+	const (
+		// rke2HelmChartNS is the namespace under which all RKE2 related HelmChart resources are deployed.
+		rke2HelmChartNS = "kube-system"
+		// namePrefix is the prefix that the core RKE2 releated HelmChart resources have.
```
frelon:
```suggestion
		// namePrefix is the prefix that the core RKE2 related HelmChart resources have.
```

internal/upgrade/reconcilers/rke2_core.go
```diff
@@ -0,0 +1,417 @@
...
+func hashContent(content string) string {
+	sum := sha256.Sum256([]byte(content))
+	return hex.EncodeToString(sum[:])
+}
+
+// findRKE2PackagedComponents locates all running RKE2 packaged components.
+// RKE2 ships its core packaged components as HelmChart resources, so this function returns a map
+// of HelmChart resources keyed by the HelmChart name.
+func findRKE2PackagedComponents(ctx context.Context, c client.Client) (map[string]helmv1.HelmChart, error) {
+	// Right now there is no conclusive way to retrieve a RKE2 packaged component by a single delimeter, so this function
```
frelon:
```suggestion
	// Right now there is no conclusive way to retrieve a RKE2 packaged component by a single delimiter, so this function
```

internal/upgrade/reconcilers/rke2_core.go
```diff
@@ -0,0 +1,417 @@
...
+	// uses a set of delimeters that are specific to RKE2 HelmChart resources. While it is possible to miss a some charts,
+	// this set of delimeters matches all HelmCharts related to the core use-case deploymet of RKE2.
```
frelon:
```suggestion
	// this set of delimeters matches all HelmCharts related to the core use-case deployment of RKE2.
```

internal/upgrade/reconcilers/kubernetes.go
```diff
@@ -0,0 +1,220 @@
...
+	Name             string `json:"name"`
+	Namespace        string `json:"namespace"`
+	ChartURL         string `json:"chartURL"`
+	ChartContentSHA  string `json:"chartContentSHA"`
+	ChartVersion     string `json:"chartVersion"`
+	ReleaseRevisions int    `json:"releaseRevisions"`
+}
+
+// PackagedComponentsSnapshot represents a snapshot of the state
+// off all packaged components related to a Kubernetes distribution.
```
frelon:
```suggestion
// of all packaged components related to a Kubernetes distribution.
```

internal/upgrade/reconcilers/rke2_core.go
```diff
@@ -0,0 +1,417 @@
...
+
+func (h *RKE2PackagedComponentsHandler) deleteSnapshotAndWait(ctx context.Context, snapshot *corev1.ConfigMap, interval time.Duration, timeout time.Duration) error {
+	// Only delete if snapshot is not already being deleted
+	if snapshot.GetDeletionTimestamp().IsZero() {
+		if err := h.Delete(ctx, snapshot); err != nil && !apierrors.IsNotFound(err) {
+			return fmt.Errorf("deleting RKE2 packaged component snapshot %q: %w", snapshot.Name, err)
+		}
+	}
+
+	return wait.PollUntilContextTimeout(ctx, interval, timeout, true, func(ctx context.Context) (bool, error) {
```
frelon, line 254: In the current config this might block for 2 minutes, would it make sense to shorten this interval? Or to requeue the reconcile after the intverval if the ConfigMap is not deleted right away?

ipetrov117: Yes, we could shorten it. Either way, if it goes into a timeout we will pick this up in the next reconciliation cycle. Still I'd prefer some sort of block in order to not prematurely schedule a reconciliation. How does a 15 second block sound to you?

frelon: Sure, let's start with 15 seconds!

ipetrov117: Ah, we are in sync, I just committed the timeout decrease. Thanks!

frelon (approved):

PR #12, opened 2026-04-27
Introduce Helm reconciliation logic

ipetrov117:
This PR introduces reconciliation logic around Helm chart upgrades.

LCM handles Helm chart upgrades by creating/updating [`HelmChart`](https://github.com/k3s-io/helm-controller/blob/master/pkg/apis/helm.cattle.io/v1/types.go#L32) resources, whose changes are then picked up by the [`Helm Controller`](https://github.com/k3s-io/helm-controller/tree/master). So the `Helm Controller` is the one that is actually doing the upgrade, while LCM is just providing the necessary upgrade information.

Covered use-cases:
| Use case | Triggers Upgrade | LCM Process |
|----------|----------|----------|
| Helm chart release missing from cluster  | No  | LCM does not deploy charts, it only upgrades them, as such if a chart is missing from the environment, LCM will mark it as `Skipped` |
| Helm release present and at desired version; missing `HelmChart`   | No  | LCM marks the chart as already at desired version   |
| Helm release present;  corresponding `HelmChart` present; both at desired version   | No  | LCM marks the chart as already at desired version  |
| Helm release present, but **not** at desired version; missing `HelmChart`  | Yes  | LCM  creates a `HelmChart` resource. `Helm Controller` picks up the new resource and creates a workload that does `helm upgrade ...` for the specified chart.  LCM waits for said workload to complete.|
| Helm release present;  corresponding `HelmChart` present; **not** at desired version   | Yes  | LCM  updates the existing `HelmChart` resource. `Helm Controller` picks up the change and creates a new workload that does `helm upgrade ...` for the specified chart. LCM waits for said workload to complete.|

General workflow notes:
1. Helm charts are always upgraded sequentially
1. Helm chart dependencies are always upgraded first
1. Helm charts defined in the `core` release manifests are always upgraded before `product` defined charts

Examples:
- Charts deployed on cluster:
  - MetalLB - `0.15.2` - deployed via `HelmChart` resource
  - Cert Manager - `1.19.2` - deployed via `helm install` only 
  - Rancher - `2.14.0-alpha13` - deployed via `helm install` only

Before upgrade:
Screenshot 2026-04-27 at 14 36 25

- Upgrading to Product Release Manifest containing charts:
   - MetalLB - `0.15.3` - coming from the referenced Core Release Manifest
   - Cert Manager - `1.19.4`
   - Rancher - `2.14.0`

After upgrade:
Screenshot 2026-04-27 at 15 03 53

- Lifecycle Manager and CRD Helm charts

### DIFF
--- a/internal/upgrade/reconcilers/helm.go
+++ b/internal/upgrade/reconcilers/helm.go
@@ -1,0 +1,468 @@
+/*
+Copyright © 2026 SUSE LLC
+SPDX-License-Identifier: Apache-2.0
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package reconcilers
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"maps"
+	"slices"
+
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/yaml"
+
+	helmv1 "github.com/k3s-io/helm-controller/pkg/apis/helm.cattle.io/v1"
+	lifecyclev1alpha1 "github.com/suse/elemental-lifecycle-manager/api/v1alpha1"
+	"github.com/suse/elemental-lifecycle-manager/internal/helm"
+	"github.com/suse/elemental-lifecycle-manager/internal/upgrade"
+	"github.com/suse/elemental/v3/pkg/manifest/api"
+)
+
+const (
+	// HelmChartNamespace is the namespace where HelmChart resources are created.
+	// The Helm Controller watches for HelmChart resources in this namespace.
+	HelmChartNamespace = "kube-system"
+)
+
+// chartUpgradeResult holds the result of a chart upgrade attempt.
+type chartUpgradeResult struct {
+	chartName string
+	state     helm.ChartState
+}
+
+// HelmReconciler reconciles HelmChart resources for Helm chart deployments.
+type HelmReconciler struct {
+	client.Client
+	helmClient helm.Client
+	// repositoryURLs caches repository name to URL mappings
+	repositoryURLs map[string]string
+	// releaseName is the name of the Release resource managing these charts
+	releaseName string
+	// releaseVersion is the target release version
+	releaseVersion string
+}
+
+// NewHelmReconciler creates a new Helm reconciler.
+func NewHelmReconciler(c client.Client, h helm.Client) *HelmReconciler {
+	return &HelmReconciler{
+		Client:         c,
+		helmClient:     h,
+		repositoryURLs: make(map[string]string),
+	}
+}
+
+func (r *HelmReconciler) Phase() upgrade.Phase {
+	return upgrade.PhaseHelmCharts
+}
+
+func (r *HelmReconciler) Reconcile(ctx context.Context, config *upgrade.Config) (*upgrade.PhaseStatus, error) {
+	if config == nil || config.HelmCharts == nil || len(config.HelmCharts.Charts) == 0 {
+		return r.Phase().SkippedStatus(), nil
+	}
+
+	return r.reconcileHelmCharts(ctx, config.ReleaseNamespacedName.Name, config.ReleaseVersion, config.HelmCharts)
+}
+
+// reconcileHelmCharts ensures the HelmChart resources exist and are up to date.
+// Only charts that are already installed on the cluster will be upgraded.
+// Charts are processed in dependency order.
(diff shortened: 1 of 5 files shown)


atanasdinov (approved):
Thank you for the detailed description! 🚀

internal/upgrade/reconcilers/helm.go
```diff
@@ -0,0 +1,468 @@
...
+// Charts are processed in dependency order.
+func (r *HelmReconciler) reconcileHelmCharts(ctx context.Context, releaseName, releaseVersion string, config *upgrade.HelmChartConfig) (*upgrade.PhaseStatus, error) {
+	logger := log.FromContext(ctx)
+
+	// Store release context for labeling HelmChart resources
+	r.releaseName = releaseName
+	r.releaseVersion = releaseVersion
+
+	// Build repository URL map for quick lookup
+	r.repositoryURLs = make(map[string]string)
```
dharmit, line 99: We are already doing this in `NewHelmReconciler`. It can be removed from here or there.

dharmit (changes requested):
Code comments are helpful, thank you! 

Just one little thing to change.

internal/upgrade/reconcilers/helm.go
```diff
@@ -0,0 +1,468 @@
...
+}
+
+// reconcileHelmCharts ensures the HelmChart resources exist and are up to date.
+// Only charts that are already installed on the cluster will be upgraded.
+// Charts are processed in dependency order.
+func (r *HelmReconciler) reconcileHelmCharts(ctx context.Context, releaseName, releaseVersion string, config *upgrade.HelmChartConfig) (*upgrade.PhaseStatus, error) {
+	logger := log.FromContext(ctx)
+
+	// Store release context for labeling HelmChart resources
+	r.releaseName = releaseName
```
frelon, line 95: It seems like this is gotten from the Config, can we store that in the struct instead of the current approach of mutating it on every reconcile?

davidcassany (approved):

internal/upgrade/reconcilers/helm.go
ipetrov117: While it is true that `repositoryURLs` is initially set in `NewHelmReconciler`, removing this line here would be a mistake and could cause a regression.

This assignment resets `repositoryURLs` on each reconcile before rebuilding it from the current upgrade configuration. That matters because the controller is a long-running process and can reconcile different `Release` resource versions over time.

For example:
1. The controller reconciles `Release` version `x.y.z`.
2. That `Release` defines release manifest with a specific set of Helm repository URLs.
3. The `Release` is updated to version `x.y.z+n`, where the release manifest specifies  a different set of repository URLs.

Without resetting the map on each reconcile, stale repository entries from the previous configuration could remain and `resolveRepositoryURL` might use an old URL for a chart repository name. That could cause a chart to be upgraded from the wrong repository, or fail altogether (e.g. because the chart version is missing from the old repo). 

Generally, in Kubernetes controllers long-running caching is not a good approach. This implementation uses `repositoryURLs` as a short-term per-reconciliation look-up in order for the multiple functions that use it to have access to all of the necessary data without having to define a lot of function parameters.

That said, in a follow up PR, we could look to tighten the struct variables, or remove them altogether and use function parameters instead. 

Please let me know whether this makes sense to you. I could also do the tightening as part of this PR if you prefer. Either way, let me know what approach works best for you.

internal/upgrade/reconcilers/helm.go
ipetrov117: Do you mean storing the whole `upgrade.Config` on the reconciler?

I avoided passing the full `upgrade.Config` into the `HelmReconciler` because it carries data unrelated to Helm reconciliation. Also, since the controller is a long-running process and may reconcile different `upgrade.Config`s over time, storing config-derived values on the reconciler still means they would need to be refreshed on each reconcile to avoid stale state.

That said, I see how these variables are confusing. In a follow-up PR we could look to tighten up the struct fields, or remove them altogether in favour of just passing a small context-style struct to the functions that actually need them.

Please let me know whether this makes sense to you. I could also do the tightening as part of this PR if you prefer. Either way, let me know what approach works best for you.

internal/upgrade/reconcilers/helm.go
dharmit: Makes sense. I don't have a preference for tightening in this or later PR.

dharmit (approved):

internal/upgrade/reconcilers/helm.go
ipetrov117: Awesome! I'll add it to my TODO list and open up a follow up PR post-0.6. Thanks!

internal/upgrade/reconcilers/helm.go
frelon: > Do you mean storing the whole `upgrade.Config` on the reconciler?

Looking at it again I would prefer to pass the needed data to where it belongs, instead of mutating the values in the struct on each `Reconcile`.

> I avoided passing the full `upgrade.Config` into the `HelmReconciler` because it carries data unrelated to Helm reconciliation. Also, since the controller is a long-running process and may reconcile different `upgrade.Config`s over time, storing config-derived values on the reconciler still means they would need to be refreshed on each reconcile to avoid stale state.

I see, but the full `upgrade.Config` is already passed to the `Reconcile` method, so this means the HelmReconciler is hiding configuration from itself.

> That said, I see how these variables are confusing. In a follow-up PR we could look to tighten up the struct fields, or remove them altogether in favour of just passing a small context-style struct to the functions that actually need them.
> 
> Please let me know whether this makes sense to you. I could also do the tightening as part of this PR if you prefer. Either way, let me know what approach works best for you.

Great, I would prefer to pass the values to where they are needed and probably wrap them in a private context struct as suggested before! 👍

ipetrov117: > Great, I would prefer to pass the values to where they are needed and probably wrap them in a private context struct as suggested before! 👍

Awesome! Are you comfortable doing the "private context structure" changes in a follow-up PR, or do you want to me to introduce them as part of this one? 

My main concern in doing them in this PR is that it would require some refactoring which would take time for something that is not a breaking change and could be done post-0.6. That said, I am happy to do them here if this is something that you think needs to happen as part of this PR.

ipetrov117: To keep things simple, I'll add it in a follow-up PR. Thanks for the feedback 👍 .

PR #19, opened 2026-05-04
Improve availability reconciliation logic

ipetrov117:
**Problem:**
On some Kubernetes upgrade runs, the  RKE2 packaged components [availability check](https://github.com/SUSE/elemental-lifecycle-manager/blob/main/internal/upgrade/reconcilers/rke2_core.go#L125) wrongfully reports the components as available, while in reality the `Helm Controller` has not yet created the new HelmChart upgrade/install Jobs. This leads to the `KubernetesUpgraded` Phase to be wrongfully marked as `Succeeded` which in turn moves the LCM to the Helm upgrade logic, while the RKE2 packaged components are not yet available.

**Reason:**
Currently, the RKE2 packaged components availability check only determines whether the active components differ from the last recorded snapshot. It does not consider whether the Kubernetes version has changed since that snapshot.

As a result, the check is not tied to the Kubernetes upgrade flow and does not wait for the `Helm Controller` to react to the new Kubernetes version. This creates a race condition where immediately after Kubernetes is upgraded, both the availability check and the `Helm Controller` are triggered, but the availability check may run before the `Helm Controller` has started upgrading the packaged components, causing the wrongful availability report.

**Solution:**
To avoid this race condition, the [PackagedComponentsSnapshot](https://github.com/SUSE/elemental-lifecycle-manager/blob/main/internal/upgrade/reconcilers/kubernetes.go#L47) has been extended to also include:
1. The Kubernetes version that the snapshotted components relate to
2. The creation time of the actual snapshot

The above points in combination with passing the [target Kubernetes version ](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/fix_race_condition/internal/upgrade/reconcilers/kubernetes.go#L65) to the availability check enables the logic to be able to correctly determine whether the components that it reconciles are actually available or the `Helm Controller` has not yet updated them.

With these changes the following use-case workflows are supported:

| Use Case | Availability Report | Notes |
|----------|----------|----------|
| Snapshot and target Kubernetes versions equal; packaged components changed  | `InProgress` | A very rare case that involves an admin manually making changes over a packaged component. Since the components have changed after the snapshot has been taken, component availability needs to be validated.  |
| Snapshot and target Kubernetes versions differ; no changed components detected; `Helm Controller` not yet scheduled a component update  | `InProgress`  | While no component checks have been detected, a change in the Kubernetes version always merits a change to at least one packaged component. Hence the availability check determines that the `Helm Controller` has not yet scheduled the component updates and marks the process as `InProgress`.  |
| Snapshot and target Kubernetes versions differ; packaged components changed; `Helm Controller` update in progress | `InProgress`  | Availability check reconciles each changed component and looks at its Job and whether it has completed after the snapshot creation. Since `Helm Controller` update is in progress, this means that not all packaged component Jobs have completed, hence marking `InProgress`  |
| Snapshot and target Kubernetes versions differ; packaged components changed; `Helm Controller` update finished  | `PackagedComponentsAvailable`  | Since the `Helm Controller` packaged components update has completed, this means that all packaged component Jobs have completed. The availability check reconciles each changed packaged component and validates that each Job is completed and more importantly has completed **after**  the initial snapshot was created. This ensures that after the snapshot all components that have changed have actually been updated by the `Helm Controller`. |
| Snapshot and target Kubernetes versions equal; no changed packaged components  |`PackagedComponentsAvailable`  |  No changed components + equal target and snapshot Kubernetes versions indicates that the state of the snapshotted components has not changed between the snapshot and the time the availability is validated. Hence components are reported as available.  |

### DIFF
--- a/internal/upgrade/reconcilers/kubernetes.go
+++ b/internal/upgrade/reconcilers/kubernetes.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"strings"
+	"time"
 
 	upgradecattlev1 "github.com/rancher/system-upgrade-controller/pkg/apis/upgrade.cattle.io/v1"
 	lifecyclev1alpha1 "github.com/suse/elemental-lifecycle-manager/api/v1alpha1"
@@ -45,6 +46,12 @@ type PackagedComponentChartSnapshot struct {
 // PackagedComponentsSnapshot represents a snapshot of the state
 // of all packaged components related to a Kubernetes distribution.
 type PackagedComponentsSnapshot struct {
+	// CreationTime is the time at which the snapshot was created.
+	CreationTime time.Time
+	// SourceKubernetesVersion is the Kubernetes version for which the snapshot was done.
+	SourceKubernetesVersion string
+	// Charts represents the charts package component state related to the given source
+	// Kubernetes version.
 	Charts []*PackagedComponentChartSnapshot
 }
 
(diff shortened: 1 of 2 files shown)


atanasdinov (approved):

davidcassany (approved):
LGTM

PR #26, opened 2026-05-14
Unit tests for upgrade reconcilers

dharmit:
This PR implements unit tests for the reconcilers under `interanl/upgrade/reconcilers`.


internal/upgrade/reconcilers/plan_test.go
```diff
@@ -0,0 +1,327 @@
...
+		})
+
+		Context("with valid label selector", func() {
+			BeforeEach(func() {
+				plan.Spec.NodeSelector = &metav1.LabelSelector{
+					MatchLabels: map[string]string{testLabelRole: testLabelWorker},
+				}
+			})
+
+			It("should return nodes matching the selector", func() {
```
davidcassany: I am not sure what this test is actually testing, feels like this is testing the label selector from k8s API.

internal/upgrade/reconcilers/plan_test.go
```diff
@@ -0,0 +1,327 @@
...
+				selector, err := metav1.LabelSelectorAsSelector(plan.Spec.NodeSelector)
+				Expect(err).NotTo(HaveOccurred())
+
+				nodes := &corev1.NodeList{}
+				err = fakeClient.List(ctx, nodes, client.MatchingLabelsSelector{Selector: selector})
+				Expect(err).NotTo(HaveOccurred())
+				Expect(nodes.Items).To(HaveLen(2))
+			})
+
+			It("should return empty list when no nodes match", func() {
```
davidcassany: Same as above, feels like this is just testing the fake client filtering by labels. At glance I'd say there is no need for us to test it, this is not something we coded.

internal/upgrade/reconcilers/plan_test.go
```diff
@@ -0,0 +1,327 @@
...
+				Expect(err).NotTo(HaveOccurred())
+
+				nodes := &corev1.NodeList{}
+				err = fakeClient.List(ctx, nodes, client.MatchingLabelsSelector{Selector: selector})
+				Expect(err).NotTo(HaveOccurred())
+				Expect(nodes.Items).To(BeEmpty())
+			})
+		})
+
+		Context("with match expressions", func() {
```
davidcassany: Same as above, unless I missed something this selection logic is not ours, so I'd say there is no need to test it.

internal/upgrade/reconcilers/os_test.go
```diff
@@ -0,0 +1,412 @@
...
+				Expect(err).NotTo(HaveOccurred())
+				Expect(status).NotTo(BeNil())
+				Expect(status.State).To(Equal(lifecyclev1alpha1.UpgradeSkipped))
+				Expect(status.Message).To(ContainSubstring("OS"))
+			})
+		})
+
+		Context("when config is nil", func() {
+			It("should skip the phase", func() {
+				status, err := reconciler.Reconcile(ctx, nil)
```
davidcassany: Out of curiosity, is this is actually a valid case? I mean is there any situation in which the reconciler gets a nil upgrade? I am wondering if it wouldn't make more sense to simply return error and force a reconcile rather skip phases and progress.
I know this more a functional question rather than a test coding concern.

internal/upgrade/reconcilers/os_test.go
```diff
@@ -0,0 +1,412 @@
...
+			It("should create only control-plane plan", func() {
+				status, err := reconciler.Reconcile(ctx, config)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(status).NotTo(BeNil())
+				Expect(status.State).To(Equal(lifecyclev1alpha1.UpgradeSucceeded))
+				Expect(planReconcileCnt).To(Equal(1)) // Only control-plane plan
+			})
+
+			It("should return succeeded when plan completes", func() {
```
davidcassany: Isn't this test running the same logic as the one above? Probably we could omit it.

internal/upgrade/reconcilers/os_test.go
```diff
@@ -0,0 +1,412 @@
...
+						Status: &upgrade.PhaseStatus{
+							State:   lifecyclev1alpha1.PlanComplete,
+							Message: testPlanCompleted,
+						},
+						Nodes: []corev1.Node{},
+					}, nil
+				}
+			})
+
+			It("should apply drain options to plans", func() {
```
davidcassany: I am not sure how this tests is actually verifying drain has been applied. It feels like it just verifies the status is set to succeed if the plan is complete.

internal/upgrade/reconcilers/os_test.go
```diff
@@ -0,0 +1,412 @@
...
+						},
+						Nodes: []corev1.Node{},
+					}, nil
+				}
+			})
+
+			It("should handle empty cluster as control-plane only", func() {
+				status, err := reconciler.Reconcile(ctx, config)
+
+				// With no nodes, it's control-plane only (single plan)
```
davidcassany: Shouldn't we verify a single plan has been created, as way to verify it goes through control plan only setup? Similarly as above I have the feeling the test does not verify the actual options. Feels like if it would have created 2 plans the test would have passed too.

internal/upgrade/reconcilers/os_test.go
```diff
@@ -0,0 +1,412 @@
...
+				Expect(status).NotTo(BeNil())
+				Expect(status.State).To(Equal(lifecyclev1alpha1.UpgradeSucceeded))
+			})
+		})
+	})
+
+	Describe("isControlPlaneOnlyCluster", func() {
+		It("should return true for all control-plane nodes", func() {
+			// This is a package-private function, but we can test it indirectly
+			// through the Reconcile behavior
```
davidcassany: This sort of logic is what I'd have expected in the previous test. Verify it is actually a control plane only cluster.

internal/upgrade/reconcilers/os_test.go
```diff
@@ -0,0 +1,412 @@
...
+				}, nil
+			}
+
+			_, err := reconciler.Reconcile(ctx, config)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(planReconcileCnt).To(Equal(2)) // Two plans: control-plane and worker
+		})
+
+		It("should return true for empty cluster", func() {
+			// No nodes created
```
davidcassany: What is this test aiming to test in comparison from the on in line 336? I'd say this one here aims to test the same but with a proper verification logic.

internal/upgrade/reconcilers/kubernetes_test.go
```diff
@@ -0,0 +1,431 @@
...
+				mockPackagedHandler.ReconcileAvailabilityFn = func(ctx context.Context, targetVersion string, snapshot *reconcilers.PackagedComponentsSnapshot) (*upgrade.PhaseStatus, error) {
+					return &upgrade.PhaseStatus{
+						State:   lifecyclev1alpha1.K8sPackagedComponentsAvailable,
+						Message: testComponentsAvailable,
+					}, nil
+				}
+			})
+
+			It("should execute control-plane plan first", func() {
+				node := testutil.NewTestNodeWithVersion("cp1", "v1.30.0", true)
```
davidcassany: Shouldn't we include a worker node here actually verify the control-plane goes first? If I understand it correctly then we would get two plans, being the control plane the first one to process. So basically the next test down below. I wonder which logic is this covering that is not already covered with the next test too.

internal/upgrade/reconcilers/kubernetes_test.go
```diff
@@ -0,0 +1,431 @@
...
+				}
+
+				_, err := reconciler.Reconcile(ctx, config)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(planNames).To(HaveLen(1))
+				Expect(planNames[0]).To(ContainSubstring("control-plane"))
+			})
+
+			It("should execute control-plane then worker plans in order", func() {
```
davidcassany: I wonder if this tests is already covering the case of the previous test.

davidcassany (commented):
Overall looks good to me, nice!
I am just having some doubts in few specific tests. I am not sure some of them are actually verifying the actual behavior and I also have the feeling some are redundant.
I am not really familiar with the code base, so it can also be that I am missing part of the context or misunderstanding the code.

internal/upgrade/reconcilers/plan_test.go
dharmit: Yes, I had the same feeling when I went through it. Should have removed it then.

internal/upgrade/reconcilers/os_test.go
dharmit: > Out of curiosity, is this is actually a valid case? I mean is there any situation in which the reconciler gets a nil upgrade? 

I agree that adding a test to validate something like `if x == nil` is like questioning/testing the Go language construct and not our business logic. To the second part of your question, I am inclined to believe it's not really going to happen (reference: [`func NewConfig`](https://github.com/SUSE/elemental-lifecycle-manager/blob/27e5da823ff32a022c36dc0014f7a71eb86e01f4/internal/upgrade/config.go#L80))

> I am wondering if it wouldn't make more sense to simply return error and force a reconcile rather skip phases and progress.

I have a different opinion: there's never going to be a scenario where the `Reconcile` method of `os` reconciler will be called with either `config` or `config.OS` being set to `nil`, unless that's by error. So we should remove that `nil` check and let the future code panic if either of these is left to `nil` and fix code when that happens.

cc @ipetrov117 need you to chime in with your thoughts. :)

internal/upgrade/reconcilers/os_test.go
dharmit: There's definitely overlap. I'll merge the two.

internal/upgrade/reconcilers/os_test.go
dharmit: True. Either me or the tool I used is hallucinating, but I remember going through more test specs for drain. Anyway, let me focus on improving it.

I can think of adding checks in the `mockPlanRec.ReconcileFn` which is in the `BeforeEach` block. Checks like:
- `Expect(desired.Spec.Drain).ToNot(BeNil())`
- `Expect(desired.Spec.Cordon).To(BeFalse())`

Also, add a test spec for the scenario when `DrainOpts` aren't provided. Checks would turn opposite of above spec in that case. WDYT?

dharmit: So, upon looking further I found that Drain related tests already exist at https://github.com/SUSE/elemental-lifecycle-manager/blob/27e5da823ff32a022c36dc0014f7a71eb86e01f4/internal/plan/base_test.go#L27

I'll rather remove them from here unless anyone thinks otherwise.

internal/upgrade/reconcilers/os_test.go
ipetrov117: The whole "skipping a phase when its corresponding config is nil" was written with the foresight that this code would be reused for CAPI [In-place updates logic](https://kubernetes.io/blog/2026/01/27/cluster-api-v1-12-release/), where the controller will act as a sort of web server and would be called by the ClusterAPI regarding whether it could do a specific task (e.g. update RKE2, OS, etc.) in that case the other upgrades would need to be skipped.

I agree that this might be thinking too much into the future and would, of course, require a more serious evaluation of the CAPI In-place logic, which due to time constraints we did not manage to do for the upcoming release. That said, I do like the flexibility that the current logic gives us and I do not think we would get to a point where it would cause problems. Especially since we are the one's that control what the `core` release manifest will hold.

IMO we should first evaluate CAPI In-place and based on what it expects decide whether the skip logic is not needed. Either way, we can reevaluate this specific approach post-release, as it is not a breaking change IMO. What do you guys think?

internal/upgrade/reconcilers/os_test.go
dharmit: Ack. Modified the test above this one. 👍🏽

internal/upgrade/reconcilers/os_test.go
dharmit: You're right. I missed this during self review. Removing.

dharmit: I have removed this one and kept the one on line 336 (it's line number has changed now), but the verification logic has been corrected. Let me know if you have other thoughts.

internal/upgrade/reconcilers/kubernetes_test.go
dharmit: Yes, you're right and I had similar thoughts when reviewing it. But forgot to remove this. IMO the next test is covering things well. So I have removed this one.

internal/upgrade/reconcilers/testutil/builders.go
```diff
@@ -0,0 +1,199 @@
...
+
+// NewTestNodeWithVersion creates a test node with a specific kubelet version.
+func NewTestNodeWithVersion(name, version string, isControlPlane bool) *corev1.Node {
+	node := NewTestNode(name, isControlPlane)
+	node.Status.NodeInfo.KubeletVersion = version
+	return node
+}
+
+// NewTestSUCPlan creates a test System Upgrade Controller Plan.
+func NewTestSUCPlan(name, namespace string) *upgradecattlev1.Plan {
```
ipetrov117, line 154: This function is only used in `plan_test.go`, wouldn't it make more sense to transfer the logic there instead of having it here?

internal/upgrade/reconcilers/testutil/builders.go
```diff
@@ -0,0 +1,199 @@
...
+	"github.com/suse/elemental-lifecycle-manager/internal/upgrade"
+	"github.com/suse/elemental/v3/pkg/manifest/api"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+// NewTestConfig creates a basic upgrade.Config for testing.
+func NewTestConfig() *upgrade.Config {
```
ipetrov117: Instead of having `NewTestConfig()`, `NewTestConfigWithOS()`, `NewTestConfigWithKubernetes()`, `NewTestConfigWithHelmCharts`, wouldn't it make more sense to have the `NewTestConfig()` support functional options? Something like:


```suggestion
type ConfigOption func(*upgrade.Config)

func NewTestConfig(opts ...ConfigOption) *upgrade.Config {
```

and then have config options like `WithOS()`, `WithKubernetes()`, `WithHelmCharts()`

This adds a layer of extensibility that would not require writing multiple different functions to construct a single `upgrade.Config` resource.

internal/upgrade/reconcilers/os_test.go
davidcassany: > The whole "skipping a phase when its corresponding config is nil"

That's not the case here if I get it right. Here all configs are nil in this test case. Just wondering what's meaning and purpose it. I can understand skipping some phase, but if the whole upgrade definition is nil I struggle to figure out how this should be progressing and even conclude. Anyway I was essentially curious about particular case, if that is intended I am fine keeping it I just couldn't understand it and wondered if we were "verifying" a misbehavior in tests.

internal/upgrade/reconcilers/plan_test.go
```diff
@@ -0,0 +1,242 @@
...
+				Expect(result.Nodes).To(BeEmpty())
+
+				// Verify plan was created in the cluster
+				created := &upgradecattlev1.Plan{}
+				err = fakeClient.Get(ctx, types.NamespacedName{
+					Name:      desiredPlan.Name,
+					Namespace: desiredPlan.Namespace,
+				}, created)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(created.Name).To(Equal("test-plan"))
```
ipetrov117: This feels redundant, we already retrieved the plan by name/namespace and now check whether the plan matches the name that we retrieved it with.

I think comparing the `spec` of the retrieved and desired plans is better here:

```suggestion
				Expect(created.Spec).To(Equal(desiredPlan.Spec))
```

internal/upgrade/reconcilers/plan_test.go
```diff
@@ -0,0 +1,242 @@
...
+			})
+		})
+
+		Context("when plan exists", func() {
+			BeforeEach(func() {
+				Expect(fakeClient.Create(ctx, desiredPlan)).To(Succeed())
+			})
+
+			It("should not recreate the plan", func() {
```
ipetrov117: I am not sure what we are testing here, the `reconciler.Reconcile` has no recreate logic. It either gets an existing plan or creates it if it does not exist. What is this test trying to cover?

internal/upgrade/reconcilers/plan_test.go
```diff
@@ -0,0 +1,242 @@
...
+				Expect(result).NotTo(BeNil())
+
+				// Verify only one plan exists
+				planList := &upgradecattlev1.PlanList{}
+				err = fakeClient.List(ctx, planList)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(planList.Items).To(HaveLen(1))
+			})
+
+			It("should parse status from existing plan", func() {
```
ipetrov117: I feel like this is already tested both when the plan is [applying](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/plan_test.go#L113) and [completed](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/plan_test.go#L130). I am not sure what we achieve from testing this also here.

internal/upgrade/reconcilers/plan_test.go
```diff
@@ -0,0 +1,242 @@
...
+				result, err := reconciler.Reconcile(ctx, desiredPlan)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(result.Status.State).To(Equal(lifecyclev1alpha1.UpgradeInProgress))
+			})
+		})
+
+		Context("when plan is applying", func() {
+			BeforeEach(func() {
+				desiredPlan.Status.Applying = []string{"node1", "node2"}
```
ipetrov117: Since the context has only one `It` block, wouldn't it make more sense to drop the `BeforeEach` and just define these lines in the `It` block?

internal/upgrade/reconcilers/plan_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+			It("should parse status from existing plan", func() {
+				result, err := reconciler.Reconcile(ctx, desiredPlan)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(result.Status.State).To(Equal(lifecyclev1alpha1.UpgradeInProgress))
+			})
+		})
+
+		Context("when plan is applying", func() {
```
ipetrov117: I feel like this context should also handle the use case where - the plan is applying on nodes, but currently the `applying` field is empty. This can happen when the plan transitions from one node to another, depending on when the polling is done, the applying list could be empty. Having a test case for this would actually cover [this](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/plan.go#L141-L143) logic .

internal/upgrade/reconcilers/plan_test.go
```diff
@@ -0,0 +1,242 @@
...
+				}
+				Expect(fakeClient.Create(ctx, desiredPlan)).To(Succeed())
+			})
+
+			It("should return complete status when no nodes match", func() {
+				result, err := reconciler.Reconcile(ctx, desiredPlan)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(result.Status.State).To(Equal(lifecyclev1alpha1.PlanComplete))
+				Expect(result.Status.Message).To(ContainSubstring("completed successfully"))
```
ipetrov117: This may be a personal preference, but I'd also check whether the message contains the plan name.

internal/upgrade/reconcilers/plan_test.go
```diff
@@ -0,0 +1,242 @@
...
+				result, err := reconciler.Reconcile(ctx, desiredPlan)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(result.Status.State).To(Equal(lifecyclev1alpha1.UpgradeFailed))
+				Expect(result.Status.Message).To(ContainSubstring("failed"))
+				Expect(result.Status.Message).To(ContainSubstring("upgrade script exited with code 1"))
+			})
+		})
+
+		Context("when plan has PlanComplete condition false without reason", func() {
```
ipetrov117: Would it make sense to have this under the "when plan is complete" context?

internal/upgrade/reconcilers/plan_test.go
```diff
@@ -0,0 +1,242 @@
...
+
+			It("should return in-progress status", func() {
+				result, err := reconciler.Reconcile(ctx, desiredPlan)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(result.Status.State).To(Equal(lifecyclev1alpha1.UpgradeInProgress))
+			})
+		})
+
+		Context("when plan has no relevant conditions", func() {
```
ipetrov117: Don't we indirectly test this in the ["when plan exists"](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/plan_test.go#L105) context? There we create a SUC Plan with an empty status and test whether we return in-progress. What do we get out of explicitly setting this context, in my eyes an empty condition is the same as an unknown condition when it comes to SUC Plan reconciliation.

davidcassany (approved):
LGTM :+1:

internal/upgrade/reconcilers/os_test.go
```diff
@@ -0,0 +1,359 @@
...
+			})
+
+			It("should create only control-plane plan and succeed", func() {
+				status, err := reconciler.Reconcile(ctx, config)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(status).NotTo(BeNil())
+				Expect(planReconcileCnt).To(Equal(1)) // Only control-plane plan
+				Expect(status.State).To(Equal(lifecyclev1alpha1.UpgradeSucceeded))
+				Expect(status.Message).To(ContainSubstring("successfully"))
```
ipetrov117: Why not match the full message string?

internal/upgrade/reconcilers/os_test.go
```diff
@@ -0,0 +1,359 @@
...
+
+				// Create worker nodes
+				worker1 := testutil.NewTestNode("worker1", false)
+				Expect(fakeClient.Create(ctx, worker1)).To(Succeed())
+
+				worker2 := testutil.NewTestNode("worker2", false)
+				Expect(fakeClient.Create(ctx, worker2)).To(Succeed())
+			})
+
+			It("should create both control-plane and worker plans in order", func() {
```
ipetrov117: Isn't this indirectly covered by the next two tests?

internal/upgrade/reconcilers/os_test.go
```diff
@@ -0,0 +1,359 @@
...
+			It("should return the failed status", func() {
+				status, err := reconciler.Reconcile(ctx, config)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(status.State).To(Equal(lifecyclev1alpha1.UpgradeFailed))
+				Expect(status.Message).To(ContainSubstring("Upgrade script failed"))
+			})
+		})
+
+		Context("with empty cluster", func() {
```
ipetrov117: I am having trouble understanding what we are testing here. If a cluster is empty (e.g. it has no nodes), then on what would LCM run let alone apply upgrades to. This feels like it should be dropped.

internal/upgrade/reconcilers/os_test.go
```diff
@@ -0,0 +1,359 @@
...
+				// With no nodes, it's control-plane only (single plan)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(status).NotTo(BeNil())
+				Expect(planReconcileCnt).To(Equal(1)) // Only one plan for control-plane only
+				Expect(status.State).To(Equal(lifecyclev1alpha1.UpgradeSucceeded))
+			})
+		})
+	})
+
+	Describe("isControlPlaneOnlyCluster", func() {
```
ipetrov117: Isn't everything under this already tested as part of the `Reconcile` tests above?

internal/upgrade/reconcilers/os_test.go
dharmit: Are you concerned with the test case covering the `nil` workflow or the business logic? It sounds like the former.

Because it's not just this file where `nil` test exists. It's also in other places (all added through this PR):
```sh
$ grep -r "is nil" internal/upgrade/reconcilers
internal/upgrade/reconcilers/helm_test.go:              Context("when HelmCharts config is nil", func() {
internal/upgrade/reconcilers/helm_test.go:              Context("when config is nil", func() {
internal/upgrade/reconcilers/kubernetes_test.go:                Context("when Kubernetes config is nil", func() {
internal/upgrade/reconcilers/kubernetes_test.go:                Context("when config is nil", func() {
internal/upgrade/reconcilers/os_test.go:                Context("when OS config is nil", func() {
internal/upgrade/reconcilers/os_test.go:                Context("when config is nil", func() {
```

internal/upgrade/reconcilers/rke2_core_test.go
```diff
@@ -0,0 +1,537 @@
...
+		findCompsFn = func(ctx context.Context, c client.Client) (map[string]helmv1.HelmChart, error) {
+			return map[string]helmv1.HelmChart{}, nil
+		}
+
+		handler = reconcilers.NewRKE2PackagedComponentsHandler(fakeClient, mockHelm, findCompsFn)
+	})
+
+	Describe("GenerateSnapshot", func() {
+		Context("when snapshot does not exist", func() {
+			It("should create new snapshot", func() {
```
ipetrov117: Would it make sense to combine this test case with the below "should populate snapshot with current component state" case? There we also create a snapshot. That way, in a single place, we would validate:
1. Whether the snapshot is created both as a struct in code and as a ConfigMap in cluster
2. Whether the snapshot holds the correct Kuberentes and chart versions.

internal/upgrade/reconcilers/rke2_core_test.go
```diff
@@ -0,0 +1,537 @@
...
+				_, err := handler.GenerateSnapshot(ctx, config)
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			It("should return existing snapshot", func() {
+				// Get the snapshot again
+				snapshot, err := handler.GenerateSnapshot(ctx, config)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(snapshot).NotTo(BeNil())
```
ipetrov117: Shouldn't we validate that what we retrieved is actually what was created beforehand?

internal/upgrade/reconcilers/rke2_core_test.go
```diff
@@ -0,0 +1,537 @@
...
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		Context("when no components changed", func() {
+			It("should return available when version matches", func() {
+				status, err := handler.ReconcileAvailability(ctx, "v1.30.0", snapshot)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(status.State).To(Equal(lifecyclev1alpha1.K8sPackagedComponentsAvailable))
+				Expect(status.Message).To(ContainSubstring("available"))
```
ipetrov117: IMO we should validate the full phase status message here. That way we can be sure that the output is what we actually expect.

internal/upgrade/reconcilers/rke2_core_test.go
```diff
@@ -0,0 +1,537 @@
...
+			})
+
+			It("should return in-progress when version differs", func() {
```
ipetrov117: I may be missing something, but I do not see a scenario where the Kubernetes version was bumped, but none of the underlying packaged components has changed. As such would it make sense to drop this test altogether?

internal/upgrade/reconcilers/rke2_core_test.go
```diff
@@ -0,0 +1,537 @@
...
+
+				//nolint:unparam
+				findCompsFn = func(_ context.Context, _ client.Client) (map[string]helmv1.HelmChart, error) {
+					return map[string]helmv1.HelmChart{testRKE2CoreDNS: *chart}, nil
+				}
+				handler = reconcilers.NewRKE2PackagedComponentsHandler(fakeClient, mockHelm, findCompsFn)
+
+				mockHelm.RetrieveReleaseFn = func(name string) (*helm.ReleaseInfo, error) {
+					return &helm.ReleaseInfo{
+						ChartVersion: testChartVersion,
```
ipetrov117: Instead of using a variable from another file, would it make sense to actually provide the values from the test chart that was created above like so:

```golang
return &helm.ReleaseInfo{
  ChartVersion: chart.Spec.Version,
  Namespace:    chart.Namespace,
```

internal/upgrade/reconcilers/rke2_core_test.go
```diff
@@ -0,0 +1,537 @@
...
+						Revisions:    1,
+					}, nil
+				}
+
+				var err error
+				snapshot, err = handler.GenerateSnapshot(ctx, config)
+				Expect(err).NotTo(HaveOccurred())
+			})
+
+			It("should wait for changed components to complete", func() {
```
ipetrov117, line 317: I am not convinced that this test is actually testing what it says it is testing..

"should wait for changed components to complete" implies that the [snapshot](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/rke2_core_test.go#L281) created in the above BeforeEach will contain the created [chart](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/rke2_core_test.go#L256), this chart would then be updated and the handler will reconcile it.

However, that is not at all what is happening.. Instead we have the following workflow:
1. In the top-level BeforeEach we [generate a snapshot](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/rke2_core_test.go#L229) with **no** helm charts related to it.
2. At the bottom, we create a chart that we assign to the [findCompFn](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/rke2_core_test.go#L266).
3. The end result is that the [GenerateSnapshot](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/rke2_core_test.go#L281) from the lowest `BeforeEach` **does not** create a new snapshot, but instead gets the already existing **empty** snapshot created from (1).

Based on the above points, this test is **not** testing changed component behaviour, but is actually testing new component behaviour, because the snapshot used for the tests has **no** charts assigned to it and the created chart is treated as a new chart. Or am I misreading the code?

internal/upgrade/reconcilers/rke2_core_test.go
```diff
@@ -0,0 +1,537 @@
...
+				Expect(fakeClient.Create(ctx, job)).To(Succeed())
+
+				status, err := handler.ReconcileAvailability(ctx, "v1.30.0", snapshot)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(status.State).To(Equal(lifecyclev1alpha1.K8sPackagedComponentsAvailable))
+			})
+		})
+
+		Context("with new components", func() {
```
ipetrov117, line 427: Shouldn't we also have a "should return available when job completes" validation, so that we are at least consistent with the above test context?

internal/upgrade/reconcilers/rke2_core_test.go
```diff
@@ -0,0 +1,537 @@
...
+				chart.Spec.ChartContent = testNewContent
+				Expect(fakeClient.Update(ctx, chart)).To(Succeed())
+
+				status, err := handler.ReconcileAvailability(ctx, "v1.30.0", snapshot)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(status.State).To(Equal(lifecyclev1alpha1.UpgradeInProgress))
+			})
+
+			It("should return available when job completes", func() {
```
ipetrov117:
```suggestion
			It("should return available Job for changed component completes", func() {
```

internal/upgrade/reconcilers/rke2_core_test.go
```diff
@@ -0,0 +1,537 @@
...
+
+				status, err := handler.ReconcileAvailability(ctx, "v1.30.0", snapshot)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(status.State).To(Equal(lifecyclev1alpha1.UpgradeInProgress))
+			})
+		})
+	})
+
+	Describe("snapshot lifecycle", func() {
```
ipetrov117: I expected to see this type of validation from the "GenerateSnapshot" block. Wouldn't it make more sense to merge "GenerateSnapshot" and "snapshot lifecycle"?

internal/upgrade/reconcilers/rke2_core_test.go
```diff
@@ -0,0 +1,537 @@
...
+			err = fakeClient.Get(ctx, types.NamespacedName{
+				Name:      testSnapshotName,
+				Namespace: config.ReleaseNamespacedName.Namespace,
+			}, cm)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(cm.Annotations["lifecycle.suse.com/source-kubernetes-version"]).To(Equal("v1.31.0"))
+		})
+	})
+
+	Describe("job completion verification", func() {
```
ipetrov117: Similar to the above comment. Shouldn't this be part of the `ReconcileAvailability` validation?

internal/upgrade/reconcilers/os_test.go
ipetrov117: IMO we should drop the "when config is nil" tests and focus only on the "when OS/K8s is nil" use-cases.

internal/upgrade/reconcilers/kubernetes_test.go
```diff
@@ -0,0 +1,409 @@
...
+		Context("when Kubernetes config is nil", func() {
+			It("should skip the phase", func() {
+				config.Kubernetes = nil
+
+				status, err := reconciler.Reconcile(ctx, config)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(status).NotTo(BeNil())
+				Expect(status.State).To(Equal(lifecyclev1alpha1.UpgradeSkipped))
+				Expect(status.Message).To(ContainSubstring("Kubernetes"))
```
ipetrov117: This may be personal preference, but IMO validating the full string whenever possible is better than validating only the substring.

internal/upgrade/reconcilers/kubernetes_test.go
```diff
@@ -0,0 +1,409 @@
...
+			})
+		})
+
+		Context("when config is nil", func() {
```
ipetrov117: I think we can drop this and focus only on the use-cases where Kubernetes is nil.

internal/upgrade/reconcilers/kubernetes_test.go
```diff
@@ -0,0 +1,409 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	testK8sVersion          = "v1.30.0"
+	testCompleteMsg         = "Complete"
+	testComponentsAvailable = "Components available"
+)
+
+var _ = Describe("KubernetesReconciler", func() {
```
ipetrov117, line 43: May have missed it, but I am not sure I am seeing control-plane only tests such as the ones that are available in the [os_test.go](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/os_test.go#L93) file. For consistency, would it make sense to add such a test here as well.

internal/upgrade/reconcilers/kubernetes_test.go
```diff
@@ -0,0 +1,409 @@
...
+
+				status, err := reconciler.Reconcile(ctx, config)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(status.State).To(Equal(lifecyclev1alpha1.UpgradeInProgress))
+				Expect(callCount).To(Equal(1))
+			})
+		})
+
+		Context("node version verification", func() {
```
ipetrov117: Isn't node version verification related to the plan's execution? Meaning that we do not move to the next plan unless the current plan's nodes are ready and available. As such wouldn't it make sense to have this under the "plan execution" section?

internal/upgrade/reconcilers/kubernetes_test.go
```diff
@@ -0,0 +1,409 @@
...
+						State:   lifecyclev1alpha1.K8sPackagedComponentsAvailable,
+						Message: testComponentsAvailable,
+					}, nil
+				}
+
+				status, err := reconciler.Reconcile(ctx, config)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(status.State).To(Equal(lifecyclev1alpha1.UpgradeSucceeded))
+				Expect(status.Message).To(ContainSubstring("successfully"))
```
ipetrov117: May be a personal preference, but whenever possible I prefer to check the full string instead of a substring. That way we can be sure that we are not accidentally reporting a weird status message.

internal/upgrade/reconcilers/helm_test.go
```diff
@@ -0,0 +1,385 @@
...
+						Namespace:    testNamespace,
+						Config:       map[string]any{},
+						Revisions:    1,
+					}, nil
+				}
+
+				status, err := reconciler.Reconcile(ctx, config)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(status).NotTo(BeNil())
```
ipetrov117, line 122: Shouldn't we actually check the reported status instead of just validating whether it is not nil? Otherwise we cannot be sure that the reconcile is successful.

internal/upgrade/reconcilers/helm_test.go
```diff
@@ -0,0 +1,385 @@
...
+				status, err := reconciler.Reconcile(ctx, config)
+
+				Expect(err).To(HaveOccurred())
+				Expect(err.Error()).To(ContainSubstring("helm client error"))
+				Expect(status).NotTo(BeNil())
+				Expect(status.State).To(Equal(lifecyclev1alpha1.UpgradeFailed))
+			})
+		})
+
+		Context("with chart already at target version", func() {
```
ipetrov117: Isn't a chart already at target version considered a valid chart? Would it make sense to merge this with the above test?

internal/upgrade/reconcilers/helm_test.go
```diff
@@ -0,0 +1,385 @@
...
+				}
+
+				status, err := reconciler.Reconcile(ctx, config)
+
+				Expect(err).NotTo(HaveOccurred())
+				Expect(status.State).To(Equal(lifecyclev1alpha1.UpgradeSucceeded))
+			})
+		})
+
+		Context("with chart needing upgrade", func() {
```
ipetrov117: Same as the above, isn't a chart in need of an upgrade a valid chart and as such shouldn't it be under the "with valid charts" context?

ipetrov117 (commented):
Hey @dharmit, thanks for doing this. 

I did an initial pass on the PR and left a few comments. I’ll plan to take another look once we’ve had a chance to discuss or address the current feedback.

Apart from the comments, we could also consider the following suggestions:
1. Right now there are some `BeforeEach` statements defined in contexts consisting of a single `It` block - while not breaking or bad, would it make sense to drop those statements and just keep the logic in the `It` blocks themselves? That way a single test case would be easier to follow. Here is an [example](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/plan_test.go#L113-L128) of what I mean.
2. We should be consistent in testing `control-plane only` vs `control-plane + worker` upgrades - right now in `os_test.go` we attempt to test [control-plane only](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/os_test.go#L93) cluster upgrades, where in `kuberentes_test.go` we skip this altogether. I think we should better align the upgrade use-cases between OS and Kubernetes upgrades (e.g. have tests for control-plane only and control-plane + worker use-cases).

internal/upgrade/reconcilers/testutil/builders.go
dharmit: Thanks for the suggestion. It'd also align with the design pattern we have been following elsewhere in code (at least I remember it in elemental).

Done.

internal/upgrade/reconcilers/plan_test.go
dharmit: Makes sense. But, for some reason, `created.Spec.NodeSelector.MatchLabels` is being set to `nil` even though that of `desirePlan` isn't. I'm struggling to figure why that's the case.

internal/upgrade/reconcilers/plan_test.go
dharmit: Removed.

internal/upgrade/reconcilers/plan_test.go
dharmit: Removed.

internal/upgrade/reconcilers/plan_test.go
dharmit: Removed such `Context`-`It` from `plant_test.go`.

internal/upgrade/reconcilers/plan_test.go
dharmit: Agree. Added.

internal/upgrade/reconcilers/plan_test.go
dharmit: Changed to match entire message.

internal/upgrade/reconcilers/plan_test.go
dharmit: I have removed most of `Context`s with a single `It`. So this is also under a common `Context` now.

internal/upgrade/reconcilers/plan_test.go
dharmit: Removed.

internal/upgrade/reconcilers/os_test.go
dharmit: Done.

internal/upgrade/reconcilers/os_test.go
dharmit: Yes, removed.

internal/upgrade/reconcilers/os_test.go
dharmit: Yes. I had a WTH reaction when I saw it. But missed removing it. My bad.

internal/upgrade/reconcilers/os_test.go
dharmit: Removed. Thanks for pointing.

internal/upgrade/reconcilers/rke2_core_test.go
dharmit: Agree. Done.

internal/upgrade/reconcilers/rke2_core_test.go
dharmit: Agree. Done.

internal/upgrade/reconcilers/rke2_core_test.go
dharmit: Yes, done.

internal/upgrade/reconcilers/rke2_core_test.go
dharmit: Yes, dropped.

internal/upgrade/reconcilers/rke2_core_test.go
dharmit: Modified in this case. But I feel that using constants from across the files & packages is pretty common. Here it made sense however because value could be fetched from a much closer avenue.

internal/upgrade/reconcilers/rke2_core_test.go
dharmit: > At the bottom, we create a chart that we assign to the [findCompFn](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/rke2_core_test.go#L266).

According to my understanding, its this `findCompFn` that hacks the behaviour such that it feels like a component has changed. Things are surely not happening the way they happen in actual code. It's more like mocking the behaviour which makes it looks like something has changed.

> Or am I misreading the code?

I am very sure you aren't because you are the one who wrote the entire thing. I might be making some mistake here, so feel free to correct me.

internal/upgrade/reconcilers/rke2_core_test.go
dharmit: Yes, added. Thanks for pointing.

internal/upgrade/reconcilers/rke2_core_test.go
dharmit: Agree. Merged `It("should handle snapshot creation and parsing", func() {` from here into the `It("should create & populate snapshot with current component state", func() {` test. And put the deletion into `GenerateSnapshot`.

internal/upgrade/reconcilers/rke2_core_test.go
dharmit: Moved.

internal/upgrade/reconcilers/kubernetes_test.go
dharmit: Agree. Just checking for one word in a string is pointless. Done.

internal/upgrade/reconcilers/kubernetes_test.go
dharmit: Done.

internal/upgrade/reconcilers/kubernetes_test.go
dharmit: It's not commented as much, but there are tests that are being run on a test cluster with only control plane nodes. Can you suggest any specific case you'd like to be added or maybe separated from how it's being tested currently?

internal/upgrade/reconcilers/kubernetes_test.go
dharmit: Done.

internal/upgrade/reconcilers/kubernetes_test.go
dharmit: Agree.

internal/upgrade/reconcilers/helm_test.go
dharmit: Agree. Done.

internal/upgrade/reconcilers/helm_test.go
dharmit: Merged above.

internal/upgrade/reconcilers/helm_test.go
dharmit: It can be. I have moved it.

internal/upgrade/reconcilers/testutil/builders.go
dharmit: Since all testing related scaffolding is under `testutil`, I'd rather prefer it stay here. No other reason.

frelon (approved):

internal/upgrade/reconcilers/rke2_core_test.go
ipetrov117: > According to my understanding, its this findCompFn that hacks the behaviour such that it feels like a component has changed. Things are surely not happening the way they happen in actual code. It's more like mocking the behaviour which makes it looks like something has changed.

I can see how this can be confusing. Let me try to explain.

A changed component is a component whose state has change in comparison with its snapshotted counterpart.

A new component is a component that is not present in the snapshot, but has been detected as K8s packaged component. For example a new component that has been included in the K8s version that we upgraded to.

The `findCompFn` logic mocks the K8s packaged components retrieval. It does not mock the "changed vs new" component logic.

The `findCompFn` is also called both when creating a [new snapshot](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/rke2_core.go#L235) (to populate the snapshot) and during [component availability reconciliation](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/rke2_core.go#L333) (to retrieve the latest components on the cluster).

Now with all those pieces, the following is happening in the tests:
1. We create an empty snapshot from a config [here](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/rke2_core_test.go#L270). The snapshot is empty as the config we pass, holds no chart data.
2. We "find" a new chart by setting it up in `findCompFn` [here](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/rke2_core_test.go#L297-L299).
3. We **get** the same snapshot from (1) [here](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/rke2_core_test.go#L312). We get the same snapshot, because we refer to the same config from (1).
4. We call the `ReconcileAvailability` logic [here](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/rke2_core_test.go#L321).
5. The logic sees an empty snapshot and one packaged component (returned from `findCompFn`). Since there is no counterpart to the found component inside the snapshot, the logic treats this component as "new", not "changed.

This is why I believe that this test is testing "new" component logic and not "changed" component logic. 

To test the "changed" component logic, you would need to pass a config with a different version/release label to the one defined in (1). Then the `GenerateSnapshot` will see that there is a mismatch between the existing snapshot and the aforementioned config and will [recreate](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/rke2_core.go#L108-L114) the snapshot. This recreate will go into the populate snapshot logic, which will call the [`findCompFn`](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/rke2_core.go#L235). This will ensure that during `GenerateSnapshot` whatever you have under (2) will end up in the snapshot. Then when you update the existing chart [here](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/rke2_core_test.go#L319) and call `ReconcileAvailability` it will treat the chart as a "changed" component and we will actually test the correct behaviour.

Let me know if I made things clearer, or just complicated it further.

internal/upgrade/reconcilers/plan_test.go
ipetrov117: IIUC, [here](https://github.com/dharmit/elemental-lifecycle-manager/blob/CORE-575/internal/upgrade/reconcilers/testutil/builders.go#L162) we create an empty map, but the K8s fake client normalises empty maps as nil, so that is why you are seeing this. 

A quick workaround would be to just set `NodeSelector: &metav1.LabelSelector{}` and move on. However, this raises a bigger question, which is that the plan should probably have some default values in spec as otherwise it is not really a valid plan, but I guess this can be tackled in a follow-up PR.

internal/upgrade/reconcilers/kubernetes_test.go
ipetrov117: Yes, I got confused in the naming between `os_test.go` and `kubernetes_test.go`. So I guess all is good.

internal/upgrade/reconcilers/plan_test.go
dharmit: > but the K8s fake client normalises empty maps as nil, so that is why you are seeing this.

Thanks, I couldn't figure this out.

I have added a label to `desiredPlan` within this test so that the two `.Spec`s could be checked for eqaulity. Let me know what you think about it.

internal/upgrade/reconcilers/rke2_core_test.go
dharmit: Makes sense! I have changed the `config.ReleaseVersion` just before `GenerateSnapshot` is getting called in the second `BeforeEach` function. Stepping through the code, I do see recreate logic getting triggered. Do you think this matches your expecation?

Thanks a lot for such an elaborate explanation. 🙏🏽

internal/upgrade/reconcilers/plan_test.go
ipetrov117: Looks good.

internal/upgrade/reconcilers/rke2_core_test.go
ipetrov117: Yep, this should now test a component change event.

ipetrov117 (approved):

PR #29, opened 2026-05-19
Add buildtags for OBS / oci publishing

dirkmueller:


dharmit (approved):

davidcassany (approved):
LGTM

Just wondering if we should establish some sort of criteria on which receipts goes into gitea and which goes into GH. As a general rule of thumb I am in favor of adding as much as possible in GH (like in this repository) as this allows to set changes in a single PR (or even set the version as placeholder from a parent tag of the same code).
However note that the version added here is also hardcoded in gitea for the image and RPM 😕 Probably a topic a design session? WDYT?

frelon (approved):

PR #36, opened 2026-05-20
Configure GitHub pages for LCM

dharmit:
Jekyll configuration has been copy-pasted from https://github.com/SUSE/elemental/pull/391.

### DIFF
--- a/.github/workflows/jekyll-github-pages.yml
+++ b/.github/workflows/jekyll-github-pages.yml
@@ -27,16 +27,16 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Setup Pages
-        uses: actions/configure-pages@45bfe0192ca1faeb007ade9deae92b16b8254a0d
+        uses: actions/configure-pages@45bfe0192ca1faeb007ade9deae92b16b8254a0d  # v6.0.0
       - name: Build with Jekyll
-        uses: actions/jekyll-build-pages@44a6e6beabd48582f863aeeb6cb2151cc1716697
+        uses: actions/jekyll-build-pages@44a6e6beabd48582f863aeeb6cb2151cc1716697 # v1.0.13
         with:
           source: ./docs
           destination: ./_site
       - name: Upload artifact
-        uses: actions/upload-pages-artifact@fc324d3547104276b827a68afc52ff2a11cc49c9
+        uses: actions/upload-pages-artifact@fc324d3547104276b827a68afc52ff2a11cc49c9  # v5.0.0
 
   # Deployment job
   deploy:
@@ -48,4 +48,4 @@ jobs:
     steps:
       - name: Deploy to GitHub Pages
         id: deployment
-        uses: actions/deploy-pages@cd2ce8fcbc39b97be8ca5fce6e763baed58fa128
+        uses: actions/deploy-pages@cd2ce8fcbc39b97be8ca5fce6e763baed58fa128 # v5.0.0
--- a/.github/workflows/jekyll-github-pages.yml
+++ b/.github/workflows/jekyll-github-pages.yml
@@ -1,0 +1,51 @@
+# Sample workflow for building and deploying a Jekyll site to GitHub Pages
+name: Deploy Jekyll with GitHub Pages dependencies preinstalled
+
+on:
+  # Runs on pushes targeting the default branch
+  push:
+    branches: ["main"]
+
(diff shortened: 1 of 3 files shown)


.github/workflows/jekyll-github-pages.yml
```diff
@@ -0,0 +1,51 @@
...
+  group: "pages"
+  cancel-in-progress: false
+
+jobs:
+  # Build job
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v6
```
ipetrov117: In order to mitigate possible supply chain attacks, we should refer to a specific SHA instead of a direct version.

docs/quickstart.md
```diff
@@ -0,0 +1,68 @@
...
+  registry: ${RELEASE_REGISTRY_URL}
+EOF
+```
+
+Where:
+
+- `${RELEASE_VERSION}` is the version of your [release manifest](https://github.com/SUSE/elemental/blob/main/docs/release-manifest.md).
+- `${RELEASE_REGISTRY_URL}` is the registry from where LCM will retrieve this manifest version
+
+For more details about the `Release` API, upgrade workflow, and monitoring or troubleshooting steps, see the [Guides](#guides) section.
```
ipetrov117: I do not think that `[Guides](#guides)` will work, as there no `Guides` section in the Quickstart document.

docs/index.md
```diff
@@ -0,0 +1,24 @@
+# Elemental Lifecycle Manager (LCM)
```
ipetrov117, line 1: While I understand we are trying to have the same(ish) content as https://suse.github.io/elemental, I do not think that the `Features` section gives much information. Furthermore, it obscures important information about the `Release` resource - although a [single line](https://github.com/SUSE/elemental-lifecycle-manager/blob/main/README.md?plain=1#L17) is obscured, it is important to have it IMO, as it paints the general picture.

Wouldn't it make sense to keep the original setup from the README.md document excluding the `Contribution` and `License` sections?

I may be biased here, but I think that `Requirements`, `Limitations` and `Quickstart` are something that the user should see immediately on the index page. I am fine splitting out the `Quickstart` into a separate document, but I really do think that `Requirements` and `Limitations` should be on the index. 

Either way, let me know what you think 🚀

.github/workflows/jekyll-github-pages.yml
dharmit: Done.

docs/quickstart.md
dharmit: Fixed.

docs/index.md
dharmit: Agree on all points. Modified.

.github/workflows/jekyll-github-pages.yml
```diff
@@ -0,0 +1,51 @@
...
+  group: "pages"
+  cancel-in-progress: false
+
+jobs:
+  # Build job
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
```
ipetrov117: Smallest of nits: can we also add a comment as to which version this SHA points to, so that we do not have to wonder in the future. You can take a look at the [golangci_lint.yml](https://github.com/SUSE/elemental-lifecycle-manager/blob/main/.github/workflows/golangci_lint.yml) and [unit_tests.yml](https://github.com/SUSE/elemental-lifecycle-manager/blob/main/.github/workflows/unit_tests.yml) files for reference.

docs/quickstart.md
```diff
@@ -0,0 +1,68 @@
+# Quickstart
+
+## Requirements
```
ipetrov117: Can we drop the requirements section from here and instead add a single line explaining that users should familiarise themselves with the LCM requirements from the index page? That way we would not have duplication of documentation. WDYT?

ipetrov117 (commented):
Great job on this! I left two small comments and will be happy to approve once we discuss them.

.github/workflows/jekyll-github-pages.yml
dharmit: Makes complete sense. Added.

docs/quickstart.md
dharmit: Agree. It's already at two places: README and index. Removed from Quickstart.

ipetrov117: I am not sure we need a separate `Requirements` section, would it make sense to have a note under Quickstart explaining this. Something like:

```
Before installing LCM, make sure you have fulfilled the requirements described in the [Requirements section](./index.md#requirements).
```

WDYT?

dharmit: Sounds good. DOne.

ipetrov117 (approved):

atanasdinov (approved):

PR #51, opened 2026-06-15
Check type for circular dependency processing

dharmit:
For a solution manifest file like one at
<https://github.com/SUSE/elemental/blob/e80bebaa4fc58db23c470c817ed3064ddc2176c8/examples/elemental/customize/single-node/suse-solution-manifest.yaml> where a chart named 'longhorn' has a dependency on a sysext of the same name, LCM's `sortChartsByDependencies` function treated both as same on the basis of name. This caused the Helm charts upgrade of LCM to fail. This PR adds a check to ensure that only a dependency of type Chart is checked for such circular dependencies.

### DIFF
--- a/internal/upgrade/reconcilers/helm_test.go
+++ b/internal/upgrade/reconcilers/helm_test.go
@@ -321,8 +321,8 @@ var _ = Describe("HelmReconciler", func() {
 
 	Describe("dependency ordering (tested indirectly)", func() {
 		It("should detect circular dependencies", func() {
-			chart1 := testutil.NewTestHelmChart("chart1", "1.0.0", testutil.WithDependencies([]string{"chart2"}))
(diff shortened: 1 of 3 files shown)


internal/upgrade/reconcilers/helm_test.go
```diff
@@ -341,11 +341,35 @@ var _ = Describe("HelmReconciler", func() {
 			Expect(err.Error()).To(ContainSubstring("circular dependency"))
 			Expect(status).NotTo(BeNil())
 			Expect(status.State).To(Equal(lifecyclev1alpha1.UpgradeFailed))
+
+		})
+
+		It("should not error when sysext with same name as the chart is a dependency", func() {
```
ipetrov117: nit:
```suggestion
		It("should not error when sysext dependency has same name as helm chart", func() {
```

frelon (approved):
Nice!

ipetrov117 (approved):
Great find! This should also be back ported to the `0.1` branch.

PR #53, opened 2026-06-19
Custom chart value updates via LCM

### LINKED ISSUE #43: Helm chart custom values update via LCM
LCM needs to be able to update the values of a Helm chart that it is upgrading.

This can be useful when a new Helm chart version introduces a new field that must be populated with user-provided data for the chart to function correctly.

ipetrov117:
closes: #43 

This PR gives a way to update the values of a Helm chart defined in the release manifest during the wider release version upgrade process.

I am not exactly happy with the approach and it can definitely be improved further in the future, but I think it offers a good baseline.

Q&As:

_1. How does the `Release` API change?_

The `Release` CR now supports passing extra release manifest component configurations through the [componentConfiguration](https://github.com/ipetrov117/elemental-lifecycle-manager/blob/custom-chart-values/api/v1alpha1/release_types.go#L97) field. 

As of this moment only `helm` chart values configuration could be passed, but this can easily be extended to also pass OS, or K8s configurations in the future. Also, in the future, having this setup would enable us to extract the `componentConfiguration` to a separate CR that could be managed by a separate reconciler loop that handles component updates separately from component upgrades, but that is not in this PR's scope. 

The `componentConfiguration` is entirely optional and will not break already existing CR APIs.

_2. How are custom values defined in the `Release` CR?_

Under `componentConfiguration.helm` users define the `chart` that they wish to configure, as seen in the release manifest.

After specifying the desired chart, there are two ways of providing the custom values:
* Through direct inline yaml definitions:
```yaml
spec:
  ...
  componentConfiguration:
    helm:
    - chart: example
      values:
        replicaCount: 1
        resources:
          requests:
            cpu: 250m
            memory: 250Mi
```
* By pointing to an already existing secret key that holds the desired value:
```yaml
# Secret
apiVersion: v1
kind: Secret
metadata:
  name: example-values-secret
  namespace: kube-system
type: Opaque
stringData:
  values-example-chart: |-
    resources:
      requests:
        cpu: 100m

# Release CR
spec:
  ...
  componentConfiguration:
    helm:
    - chart: example
      valuesFrom:
        secretRef:
          name: example-values-secret
          keys:
          - "values-example-chart"
```

Both `helm.values` and `helm.valuesFrom` can be used together, where `helm.valuesFrom` always takes precedence over `helm.values`. This behaviour is enforced directly by the Helm Controller, not LCM. LCM just passes `helm.values`  to `spec.valuesContent` and `helm.valuesFrom.secretRef` to `spec.valuesSecrets` of the `HelmChart` resource.

_3. Why do we need both `helm.valuesFrom` and `helm.values`?_

While `helm.values` is generally preferred for small configurations and simple charts, there are cases where it is not the best fit. For example:

* Large chart configurations spanning multiple chart resources. Managing these through `helm.values` can significantly bloat the `Release` CR.
* Sensitive chart configuration, such as tokens, users, credentials, or similar values.

For these use cases, `helm.valuesFrom.secretRef` is a better fit. It works well for both sensitive data and large chart configurations. For example, a single Secret can contain multiple chart value keys, which can then be referenced by different charts within the Release CR.

_4. Why does the `helm.valuesFrom.secretRef` need to refer to a secret in the `kube-system` namespace?_

`HelmChart` resources are deployed inside of the `kube-system` namespace, as that is the namespace that the Helm Controller is managing. 

`helm.valuesFrom.secretRef` maps to the `spec.valuesSecrets` property of the `HelmChart` resource. As such, in order for the `Helm Controller` to be able to locate the secret, this secret needs to be deployed under the same namespace as the `HelmChart` resource -> `kube-system`.

I thought about the implications of this requirement (not great UX, user is required to have access to `kube-system`, etc.), however the tradeoffs seem to far outweigh any custom implementation on our side. 

One flow I considered was having the user define the secret on LCM namespace level and have LCM transfer the secret, but this opens us up to significant corner-case problems and collision issues. So I decided to take the safest approach and improve when we have an actual user use-case.

_5. How are chart value contents merged?_

Value contents are merged in the following order, from least to greatest precedence:
1. Chart default values.
2. Chart custom values defined during the initial chart deployment (e.g. during `elemental3 customize`, or a standard `helm install`.
3. Chart values defined in the target release manifest.
4. Chart values defined in `helm.values`.
5. Chart values defined in `helm.valuesFrom`.

> Note: The value precedence is not enforced by LCM, but by the Helm Controller itself. LCM just passes the values to the necessary fields.

_6. When can chart values be updated?_

As of this moment, chart values will only be updated during the broader release version upgrade process. Updating a single chart value without incrementing the release version is not yet supported.

I opted for this approach, as the broader "How do we tackle component updates?" question has not been answered **yet**.

_7. Can I edit the values of a chart without upgrading its version?_

Yes, as long as the broader release version is updated as well.

_8. Which scenarios were tested?_

Scenarios tested:
* Upgrade chart without any custom values defined either in the release manifest or `Release` CR.
* Upgrade chart version and update chart values with values coming from:
   * Release `helm.values`
   * Release `helm.valuesFrom`
   * Release manifest
   * Release manifest + Release `helm.values`
   * Release manifest + Release `helm.valuesFrom`
   * Release manifest + Release `helm.values` + Release `helm.valuesFrom`


internal/controller/release_controller.go
```diff
@@ -227,13 +227,26 @@ func (r *ReleaseReconciler) parseDrainOpts(ctx context.Context, release *lifecyc
 }
 
 func (r *ReleaseReconciler) parseUpgradeConfig(ctx context.Context, manifest *resolver.ResolvedManifest, release *lifecyclev1alpha1.Release) (config *upgrade.Config, err error) {
+	componentConfig := release.Spec.ComponentConfig
```
dharmit: Nit: can we have this closer to its first call at `if componentConfig != nil {`?

internal/controller/release_controller.go
```diff
@@ -227,13 +227,26 @@ func (r *ReleaseReconciler) parseDrainOpts(ctx context.Context, release *lifecyc
...
 	if err != nil {
 		return nil, fmt.Errorf("parsing drain options: %w", err)
 	}
 
-	runtimeConfig := &upgrade.RuntimeConfig{
-		DrainOpts: opts,
+	runtimeConfig := &upgrade.RuntimeConfig{DrainOpts: opts}
+
+	if componentConfig != nil {
+		if len(componentConfig.Helm) > 0 {
```
dharmit, line 239: Why not combine these two into single `if`?

internal/upgrade/reconcilers/helm_test.go
```diff
@@ -348,25 +358,40 @@ var _ = Describe("HelmReconciler", func() {
...
+					// Ensure custom values defined in release manifest and at runtime are present.
 					Expect(helmChart.Spec.Values.Raw).To(MatchJSON(expectedMergedValues.Raw))
+
+					// Ensure custom values secret defined at runtime is present.
 					Expect(helmChart.Spec.ValuesSecrets).To(Equal([]helmv1.SecretSpec{customValueSecret}))
 				})
 
-				It("should update existing HelmChart CR", func() {
+				It("should update existing HelmChart resource when it is present", func() {
+					// Expected chart values when runtimve and release values are merged together.
```
dharmit:
```suggestion
					// Expected chart values when runtime and release values are merged together.
```

docs/release-api.md
```diff
@@ -23,13 +23,46 @@ spec:
...
+#### Helm
+
+> NOTE: The `[*]` prefix refers to an entry in the `spec.componentConfiguration.helm` configuration list.
+
+| Field                           | Description                                                              | Required |
+|---------------------------------|--------------------------------------------------------------------------|:--------:|
+| `[*].chart`                     | Reference to the chart in the release that will be configured.           | `true`   |
+| `[*].values`                    | Custom Helm values configuration provided inline.                        | `false`  |
+| `[*].valuesFrom.secretRef`      | Apply Helm values configuration provided from a Kubernetes Secret.       | `false`  |
+| `[*].valuesFrom.secretRef.name` | Name of the secret from which to retrieve custom helm chart values. **The secret must be located in the `kube-system` namesapce.** | `true` |
```
dharmit:
```suggestion
| `[*].valuesFrom.secretRef.name` | Name of the secret from which to retrieve custom helm chart values. **The secret must be located in the `kube-system` namespace.** | `true` |

dharmit (approved):
Minor suggestions.

internal/controller/release_controller.go
ipetrov117: While right now it can be in a single `if`, I opted for this structure mainly looking to the future when we have OS and K8s config. With the suggested structure any new configuration will just be added under the general `if componentConfig != nil`.

davidcassany (approved):

atanasdinov (approved):

PR #59, opened 2026-07-14
Fail for unknown helm component configurations

ipetrov117:
Currently LCM permits defining helm chart components that are missing from the specified release manifest, which leads to harder to detect user configuration errors.

The suggested changes ensure that any helm charts defined under `componentConfiguration.helm` exist in the specified release manifest. 

Charts defined by the user, but missing from the release manifest are treated as unknown charts and are reported under the `Applied` Release status condition. If any unknown charts are found, the upgrade process is stopped until the user mitigates the misconfigurations. This ensures that all configurations are correct before starting the actual upgrade process.

**Example:**
```yaml
# Release resource referencing a release manifest that 
# ships metallb, endpoint-copier-operator and cert-manager, 
# but user has misdefined the metallb and cert-manager charts.
...
spec:
  ...
  componentConfiguration:
    helm:
    - chart: metalbl
      ...
    - chart: endpoint-copier-operator
      ...
    - chart: cert-managr
      ...
status:
  conditions:
  - lastTransitionTime: "2026-07-14T09:11:02Z"
    message: Release manifest retrieved successfully
    observedGeneration: 1
    reason: Succeeded
    status: "True"
    type: ManifestResolved
  - lastTransitionTime: "2026-07-14T09:11:02Z"
    message: Waiting for previous phases to complete
    observedGeneration: 1
    reason: Pending
    status: "False"
    type: OSUpgraded
  ...
  - lastTransitionTime: "2026-07-14T09:11:02Z"
    message: 'Unknown chart references: metalbl, cert-managr'
    observedGeneration: 1
    reason: Failed
    status: "False"
    type: Applied
```

### DIFF
--- a/internal/upgrade/config.go
+++ b/internal/upgrade/config.go
@@ -19,6 +19,9 @@ package upgrade
 
 import (
 	"fmt"
+	"maps"
+	"slices"
+	"strings"
 
 	"github.com/google/go-containerregistry/pkg/name"
 	"github.com/suse/elemental/v3/pkg/manifest/api"
@@ -85,6 +88,11 @@ type RuntimeConfig struct {
 	HelmCharts map[string]RuntimeHelmChartConfig
 }
 
+// RuntimeConfigError specifies an error caused due to invalid runtime configuration.
+type RuntimeConfigError string
+
+func (c RuntimeConfigError) Error() string { return string(c) }
+
 type RuntimeHelmChartConfig struct {
 	// Values specifies custom values provided by the user inline.
 	Values *apiextensionsv1.JSON
@@ -135,19 +143,27 @@ func NewConfig(manifest *resolver.ResolvedManifest, releaseVersion string, relea
 		DrainOpts: runtimeConfig.DrainOpts,
 	}
 
-	if manifest.SolutionExtension == nil {
-		config.HelmCharts = helmChartConfig(core.Components.Helm, nil, runtimeConfig.HelmCharts)
-	} else {
-		solution := manifest.SolutionExtension
-		config.HelmCharts = helmChartConfig(core.Components.Helm, solution.Components.Helm, runtimeConfig.HelmCharts)
+	var solutionCharts *api.Helm
+	if manifest.SolutionExtension != nil {
+		solutionCharts = manifest.SolutionExtension.Components.Helm
 	}
 
+	charts, err := helmChartConfig(core.Components.Helm, solutionCharts, runtimeConfig.HelmCharts)
+	if err != nil {
+		return nil, fmt.Errorf("parsing helm chart configuration: %w", err)
+	}
+
+	config.HelmCharts = charts
 	return config, nil
 }
 
-// helmChartConfig merges Helm configurations from core and solution manifests.
-func helmChartConfig(core, solution *api.Helm, runtimeConfigs map[string]RuntimeHelmChartConfig) []*HelmChartConfig {
+// helmChartConfig merges Helm configurations from the core and solution manifests with any runtime-defined overrides.
+func helmChartConfig(core, solution *api.Helm, runtimeConfigs map[string]RuntimeHelmChartConfig) ([]*HelmChartConfig, error) {
 	chartConfig := []*HelmChartConfig{}
+	unprocessedRuntimeCharts := make(map[string]struct{}, len(runtimeConfigs))
+	for chart := range runtimeConfigs {
+		unprocessedRuntimeCharts[chart] = struct{}{}
+	}
 
 	// Add core charts and repositories
 	if core != nil {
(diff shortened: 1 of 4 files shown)


internal/upgrade/config.go
```diff
@@ -135,19 +143,27 @@ func NewConfig(manifest *resolver.ResolvedManifest, releaseVersion string, relea
...
+	charts, err := helmChartConfig(core.Components.Helm, solutionCharts, runtimeConfig.HelmCharts)
+	if err != nil {
+		return nil, fmt.Errorf("parsing helm chart configuration: %w", err)
 	}
 
+	config.HelmCharts = charts
 	return config, nil
 }
 
 // helmChartConfig merges Helm configurations from core and solution manifests.
```
dharmit: Perhaps modify the comment to reflect the changed signature and scenario in which this function would error?

dharmit (approved):

internal/upgrade/config.go
ipetrov117: Expanded the comment to include that we also merge runtime provided Helm overrides 👍

dharmit:
@ipetrov117 thinking out loud here.

If a user adds a Helm chart in the updated SRM so that its lifecycle is managed through elemental workflow, wouldn't this check prevent them from doing so? 🤔

ipetrov117:
> If a user adds a Helm chart in the updated SRM so that its lifecycle is managed through elemental workflow, wouldn't this check prevent them from doing so? 🤔

@dharmit, not sure I am following. When would this "Helm chart add" take place? LCM has the full chart list as defined in both the solution and core manifests and it does the validation against it. Once a release manifest is created it is not meant to be updated with new charts, that is why we have release manifest versions in the first place. What exact use-case are you envisioning?

dharmit:
> What exact use-case are you envisioning?

There's maybe a piece of our various manifest files that I'm not understanding. But the scenario I'm imagining is:
1. An SRM is used to build an initial raw image for OS installation.
2. At a later point, user wants to add a Helm chart to their setup which was not a part of the initial SRM. And they'd like to ensure its upgrade process is tracked using LCM.
3. If they add it to the release manifest and create `Release` resource pointing to it, would the code in this PR not prevent such a Helm chart from being installed?

Now that I have written this down, I find gaps in my understanding already! LCM isn't going to upgrade unless it finds a Helm chart installed. So the scenario is very likely wrong.

And that makes me wonder if installing such a chart that didn't exist in the initial SRM using `helm` CLI and tracking its upgrade using a newer release manifest is a valid scenario? I already did this once in the past, but just curious if users could do this and then leverage LCM for the chart's lifecycle.

PS: I'm using the words "user" and "consumer" interchangeably here because the difference between the two in context of Elemental3 is not clear to me.

ipetrov117:
> PS: I'm using the words "user" and "consumer" interchangeably here because the difference between the two in context of Elemental3 is not clear to me.

A user is someone who uses a solution/core release manifest - this is the person that customises an image, sets up the environment based on that image and then eventually triggers an upgrade through a `Release` resource.

A consumer is a product solution that provides a solution release manifest. We call them consumers as they "consume" the core release manifest in order to produce a solution manifest.

> At a later point, user wants to add a Helm chart to their setup which was not a part of the initial SRM. And they'd like to ensure its upgrade process is tracked using LCM.

Here are you referring to a user-defined Helm chart, or a SRM defined Helm chart? 

For user-defined charts, LCM is not designed to track and manage those. LCM is strictly responsible for upgrading the components that construct the underlying environment that the user is using (e.g. whatever is defined in the release manifest). Think of it as an infrastructure management tool. If we mix in its responsibilities management of user charts then the upgrade process will become highly unpredictable, as we would have no way of knowing what the user might deploy and how it might react in our upgrade workflow. That is why we handle the infrastructure part of the upgrade (e.g. OS, K8s, core/solution charts) and let the user handle their own chart upgrades.

For a newly defined chart in a SRM, this becomes tricky, as LCM does not yet support deploying Helm charts from a release manifest. In theory a user could manually deploy the desired SRM chart using `helm install` and then trigger an upgrade which would lead LCM to create a `HelmChart` resource and start managing this chart, but this is a very dirty workaround. It is dirty, as we have no guarantee that the user will deploy the correct thing with the correct configurations. Ultimately this would lead to a lot of unforeseen errors. This is a temporary LCM constraint that will be soon mitigated, as "deploying charts from a release manifest" is already on the LCM roadmap.

> If they add it to the release manifest and create Release resource pointing to it, would the code in this PR not prevent such a Helm chart from being installed?

So the chart will not be installed, yes, but not because of this suggested code, but because LCM in general does not support installing charts from a release manifest **yet**. As I mentioned this is a temporary constraint that will soon be mitigated.

> LCM isn't going to upgrade unless it finds a Helm chart installed.

Correct, right now LCM only upgrades charts it finds on the cluster, any other charts are skipped.

> And that makes me wonder if installing such a chart that didn't exist in the initial SRM using helm CLI and tracking its upgrade using a newer release manifest is a valid scenario? 

It is possible to do this, yes, but as I said this is extremely dirty and not advisable. But if someone really has to do it, they could manually install the chart and have LCM pick it up on the next upgrade cycle. But again, we should not advertise this, IMO this is a last resort scenario and we should look forward to introducing this logic in LCM.

dharmit:
My questions weren't really differentiating between user-defined chart and charts added later in an SRM/CRM. But your responses covered it all. This one is a great summary, and I'd prefer we call this out explicitly in the docs, if not already:

> Think of it as an infrastructure management tool. If we mix in its responsibilities management of user charts then the upgrade process will become highly unpredictable, as we would have no way of knowing what the user might deploy and how it might react in our upgrade workflow.

davidcassany (approved):

PR #97, opened 2026-09-02
Enables LCM auto-upgrade

dharmit:
A separate phase for LCM upgrade is added as the first phase for LCM so that it gets upgraded before anything else.


internal/upgrade/reconcilers/lcm_test.go
```diff
@@ -202,6 +204,78 @@ var _ = Describe("LCMReconciler", func() {
 				Expect(status.State).To(Equal(lifecyclev1alpha1.UpgradeSucceeded))
 				Expect(status.Message).To(Equal("All 2 LCM charts upgraded successfully (0 skipped)"))
 			})
+
+			FIt("should not error even if the charts fail to upgrade", func() {
```
dharmit: Created this for the purpose of discussion. In this test, CRD charts fail to upgrade, but that doesn't prevent LCM chart upgrades from getting triggered. 

Do we want to keep this behaviour or break if the LCM CRD chart fails to upgrade? 

Another question: do we want to return an error if either of the LCM CRD or LCM charts fail to upgrade so that the next phase doesn't get triggered? IIUC, that's not the case presently and we return a `nil` in error which leaves it upon the user to figure what went wrong with chart upgrades. I understand why this is OK in case of other Helm charts that are being upgraded towards the end but, for LCM charts, what behaviour do we think is more appropriate?

dharmit: Sharing my own thoughts. I personally feel we should not attempt to upgrade LCM chart if its CRDs fail to upgrade. And for the second one, assuming I understand things correctly, I feel we should stop the overall upgrade instead of letting OS upgrade phase get triggered.

dharmit: I have modified the PR to include the implementation for failure on LCM chart upgrade. It's in a separate commit to make removal easier if we decide not to fail the overall upgrade when one of these charts' upgrades fails.

internal/upgrade/reconcilers/helm.go
```diff
@@ -97,7 +97,15 @@ func (r *HelmReconciler) reconcileHelmCharts(ctx context.Context, releaseName, r
 	r.releaseName = releaseName
 	r.releaseVersion = releaseVersion
 
-	orderedChartConfigs, err := sortChartConfigsByDependencies(chartConfigs)
+	var workloadCharts []*upgrade.HelmChartConfig
+	for _, chartCfg := range chartConfigs {
+		name := chartCfg.Chart.GetName()
+		if !isLCMChart(name) {
```
davidcassany: Do we actually need to prevent LCMchart to be processed here? Up to my understanding it should not hurt attempting to upgrade it again, the upgrade if it is already at the desired version should do nothing.
If so, then we can probably keep the helm reconciler unmodified and if we ever set an upgrade pipe without LCMreconciler the helm reconciler would still run the full upgrade.

internal/upgrade/reconcilers/lcm.go
```diff
@@ -0,0 +1,173 @@
...
+
+	var failed, inProgress, succeeded, skipped int
+	var failedChart string
+
+	for _, result := range results {
+		switch result.state {
+		case helm.ChartStateFailed:
+			failed++
+			if failedChart == "" {
+				failedChart = result.chartName
```
davidcassany: nit: this is just keeping the last failed chart and the number of failures is not user any where. Probably you could simply directly return here with: 

```
return &upgrade.PhaseStatus{
    State:   lifecyclev1alpha1.UpgradeFailed,
	Message: fmt.Sprintf("Chart %s upgrade failed", failedChart),
}
```

without the need to count the number of failures and then check if it is bigger than zero.

davidcassany (approved):
Looks good to me

Just a couple of minor comments which do not necessarily imply any requested change, they are just arguable and minor improvements.

internal/upgrade/reconcilers/lcm_test.go
davidcassany: I fully agree, any failure in that state should stop the whole pipe and report the error IMHO.

internal/upgrade/reconcilers/helm.go
dharmit: I don't see any technical issue in removing `isLCMChart` check from here. Except it being a brain thing where it feels weird to keep it in spite of having a dedicated phase for it earlier in the flow.

> if we ever set an upgrade pipe without LCMreconciler the helm reconciler would still run the full upgrade.

If we do this, it would be for the purpose of prevent LCM charts' upgrade, right? But removing `isLCMChart` check from Helm reconciler would still cause the LCM charts' upgrade. Just presenting a counter point.

I'm fine by removing the check from here if it makes more sense. Keeping this open for others' feedback/suggestions.

internal/upgrade/reconcilers/lcm.go
dharmit: This function was copy-pasted from Helm reconciler. In LCM reconciler, I don't see a reason to have this check and keep a counter at all if we agree to failing the entire upgrade if any of the LCM charts fails to upgrade.

internal/upgrade/reconcilers/helm.go
```diff
@@ -405,7 +413,7 @@ func (r *HelmReconciler) evaluateHelmChartJobStatus(ctx context.Context, chart *
 }
 
 // aggregateResults aggregates chart upgrade results into a single PhaseStatus.
-func (r *HelmReconciler) aggregateResults(results []chartUpgradeResult, totalCharts int) *upgrade.PhaseStatus {
```
ipetrov117: Why do we unbound this from the `HelmController`? While it does not use any values from it, IMO bounding it shows the intention that this function is supposed to be called only from the `HelmController` itself and not be used anywhere else without it.

internal/upgrade/reconcilers/lcm.go
```diff
@@ -0,0 +1,160 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+const (
+	ElementalLifecycleManagerChart     = "elemental-lifecycle-manager"
+	ElementalLifecycleManagerCRDsChart = "elemental-lifecycle-manager-crds"
+)
+
+type LCMReconciler struct {
```
ipetrov117: I am not completely convinced that we need a separate reconciler for LCM. Right now this reconciler is a wrapper over the `HelmController` so that it can call `reconcileChart` and immediately fail on chart failure.

I wonder, can't we just extend the `HelmController` to support a failure policy (e.g. `lenient` and `strict`), where `lenient` will be the current logic and `strict` will fail immediately on helm chart failure. Then we can have two `HelmController` instances:
1. For the LCM itself - configured with failure policy `strict`; defined before any RM / upgrade config parsing. Will be passed an upgrade config that only has the two LCM charts that need to be upgraded.
2. For all other Helm charts - This is the already existing `HelmController` with a failure policy defined as `lenient`. This controller would also be passed an upgrade config that does not hold any LCM chart details, as those would be handled by the first controller earlier.

I think this is worth considering mainly because:
1. It reuses the existing logic without defining a separate wrapper reconciler, which reduces the testing matrix and code included.
2. It fits very nicely with what we have discussed during our upgrade recovery discussion sessions - this would tie to the users configuring their recovery mode and whether they want an immediate recovery, or want a dirty cluster. I imagine for immediate recovery we would have both `HelmControllers` configured with a `strict` failure policy. On the other hand dirty clusters would required `HelmControllers` configured with `lenient` policy.

Do you see value in this?

ipetrov117 (changes requested):
Thanks for looking into this @dharmit! 

I am not sure that the suggested changes here will be enough to handle a lifecycle manager self-upgrade.

Apologies for the huge writeup below, but the "TL;DR" is that LCM is not upgraded soon enough in the reconciliation chain.

TL;DR:
With these changes the LCM upgrade logic resides in a separate reconciler under the [pipeline for component upgrades](https://github.com/dharmit/elemental-lifecycle-manager/blob/core-644/cmd/main.go#L208-L209). While, this can be an acceptable approach for a non-LCM chart, for LCM itself doing the upgrade at the component level is too late, as this means that before upgrading LCM we have parsed both the [release manifest](https://github.com/dharmit/elemental-lifecycle-manager/blob/core-644/internal/controller/release_controller.go#L117) and [component upgrade config](https://github.com/dharmit/elemental-lifecycle-manager/blob/core-644/internal/controller/release_controller.go#L127) using the older LCM version. 

This opens us up to potential issues, where, for example:
- A newer release manifest cannot be parsed by the older LCM version - will result in a broken upgrade.
- An older LCM version constructs a broken upgrade config struct - will result in an unpredictable upgrade, as vital configurations might be missing.

To ensure we do not hit the above issue, I suggest that we do the LCM upgrade first before both release manifest and upgrade config parsing. This, of course, opens up the chicken and egg problem - in order to know whether we need to upgrade LCM, we need the release manifest, but we cannot parse it with the existing logic as that would open us up to the very problem we wish to avoid. So how can we approach this? From my PoV (and this is just a suggestion), we could:
1. Extend the way we read manifests under [`sourceReader`](https://github.com/dharmit/elemental-lifecycle-manager/blob/core-644/internal/release/manifest.go#L39) - we could have a `ReadURI()` function that reads the manifest from a URI and a `Read()` function that reuses `ReadURI()` but accepts a `source.ReleaseManifestSource` as a parameter. That way we would not be bound by referring to the `github.com/suse/elemental/v3/pkg/manifest` package, hence not bound by the library being older, while also keeping manifest resolution compatibility.
2. Have a static definition of what we expect that every manifest will have at a minimum - this will define something that we know every chart will have (e.g. name, version, repository, values), even between versions and use that to locate the LCM charts. This is fine, as if we decide to remove name, version, repository or values this would be a breaking change way before it gets to LCM.
3. Use the definition from (2) and manifest `[]byte` retrieval logic from (1) to locate the LCM charts - the important bit here is that we will not parse the full manifest, but just a portion of it that we expect to be there. This will dramatically reduce the failure vector and ensure compatibility between versions. 
4. Now that we have the LCM charts information, we can proceed to upgrade the LCM chart before we pass any final manfiest / upgrade config (how we upgrade LCM is a different topic suggested in a different comment)

By doing this we will ensure that the upgraded LCM version will be the one that does the actual parsing of the manifest and upgrade config. That said there may be better approaches to achieving our goal, this was just off the top of my mind.

dharmit:
Thanks a lot for taking time to share detailed review, @ipetrov117! 🙏🏽 

I could be misunderstanding your suggestions/concerns, so please correct me wherever required.

> * A newer release manifest cannot be parsed by the older LCM version - will result in a broken upgrade.

IIUC, this would cause a problem even if we move the upgrade logic to a different place in the flow. Instead, IMO, we should first upgrade the LCM using the old/existing logic such that it's able to parse old as well as new versions of manifest!

> * An older LCM version constructs a broken upgrade config struct - will result in an unpredictable upgrade, as vital configurations might be missing.

Isn't this something that should be handled at tests level? That is, we should be testing things with as many incorrect/broken upgrade config structs as we can.



> 2\. Have a static definition of what we expect that every manifest will have at a minimum - this will define something that we know every chart will have (e.g. name, version, repository, values), even between versions and use that to locate the LCM charts. This is fine, as if we decide to remove name, version, repository or values this would be a breaking change way before it gets to LCM.

And what if we want/need to change the static definition at some point?

> 3\. the important bit here is that we will not parse the full manifest, but just a portion of it that we expect to be there. This will dramatically reduce the failure vector and ensure compatibility between versions.

And if this specific portion of the manifest changes? Also, the versions of release manifest not staying compatible is exactly what we are basing this discussion on, no? I'm basing these two statements on below point:

> A newer release manifest cannot be parsed by the older LCM version - will result in a broken upgrade.

My understanding (happy to be corrected): 
1. I feel that if we introduce complete incompatibility between release manifests, neither currently proposed approach nor the one you are suggesting will help us do an upgrade.
2. If release manifest versions are expected to introduce breaking change, LCM should be first updated to understand/parse the newer versions so that when release manifest breaking change is rolled out, things don't fall apart.
3. I believe "LCM self-upgrade" means LCM updating things using one of the three kinds of reconcilers in the Pipeline. Custom logic to upgrade itself can't be called "self-upgrade".
    * And it's fine if we go down that route, just saying that it's not self-upgrade.

atanasdinov:
Still haven't reviewed it myself but just chiming in.

> If release manifest versions are expected to introduce breaking change, LCM should be first updated to understand/parse the newer versions so that when release manifest breaking change is rolled out, things don't fall apart.

The goal of the implementation is for the Lifecycle Manager to upgrade itself. The statement above basically tells us that whenever there's incompatibility, we'd need to manually `helm upgrade` which is not desired.

> I believe "LCM self-upgrade" means LCM updating things using one of the three kinds of reconcilers in the Pipeline. Custom logic to upgrade itself can't be called "self-upgrade".

Self-upgrade should be taken as literal as possible -- that is, the LCM executing an operation to go from the current version to a newer one. What implementation is being used has no effect on the requirement.

internal/upgrade/reconcilers/helm.go
dharmit: > Why do we unbound this from the HelmController? While it does not use any values from it IMO bounding it shows the intention that this function is supposed to be called only from the HelmController itself and not be used anywhere else without it.

Sure, I'll revert it.

internal/upgrade/reconcilers/lcm.go
dharmit: > It fits very nicely with what we have discussed during our upgrade recovery discussion sessions - this would tie to the users configuring their recovery mode and whether they want an immediate recovery, or want a dirty cluster. I imagine for immediate recovery we would have both HelmControllers configured with a strict failure policy. On the other hand dirty clusters would required HelmControllers configured with lenient policy.

Makes sense! This could likely simplify things. I can modify the PR.

> On the other hand dirty clusters would required HelmControllers configured with lenient policy.

Not a topic for this PR, but even for dirty clusters configuration, we should stick to being strict for LCM charts because that's Elemental project requirement, not user's.

ipetrov117:
> IIUC, this would cause a problem even if we move the upgrade logic to a different place in the flow. 

Moving the upgrade logic before parsing the release manifest would fix this exact issue, as we would only parse the full release manifest once we have upgraded LCM itself first. 

> Instead, IMO, we should first upgrade the LCM using the old/existing logic such that it's able to parse old as well as new versions of manifest!

First, I absolutely agree that we should be using the existing LCM logic to do the actual LCM upgrade, that said I do not agree that we should be parsing the full manifest using the old LCM version. We cannot expect that a single LCM version can handle new RM schemas that were introduced after it was released. 

With what you suggest we would need to introduce a back port for LCM that supports the new RM schema. This would mean that an environment having LCM `0.1.0` that supports schema `v0` should be first upgraded to `0.1.1`, just so that the user could run it against the newly desired RM that will move the environment to LCM `0.2.0`. To me this defeats the purpose of having a "self-upgrading LCM". 

As a last note, I'd say that it might not always be desirable for us to back port a new RM schema to a previous version. First, because this is not a bug fix, but an actual feature. Second, because it would exponentially expand our testing matrix. Third, because we might introduce a serious breaking change that cannot "just" be handled in-code. That is my gut feeling on this precise topic, I am happy to be wrong here.

> An older LCM version constructs a broken upgrade config struct - will result in an unpredictable upgrade, as vital configurations might be missing.
> > Isn't this something that should be handled at tests level? That is, we should be testing things with as many incorrect/broken upgrade config structs as we can.

Think about the following scenario:
1. LCM `X.Y.Z` expects that the upgrade config will hold a field `foo` with a value of either `bar/baz`.
2. LCM `X.Y.Z+1` expects that the upgrade config  will hold a field `foo` with a value of either `bar/baz/pippo`.
3. We attempt to parse the upgrade config using LCM `X.Y.Z` and it immediately fails as it does not recognise the `pippo` field.

Sure, this would fail on the  SV tests that we have and would immediately showcase that parsing the upgrade config using the old LCM version is probably not the best approach. What I am trying to do here, is tackle the problem before we even get to the tests, so that once we have such a scenario as described above, we handle it correctly and not have to scramble for a solution at the last moment.

> 2. Have a static definition of what we expect that every manifest will have at a minimum - this will define something that we know every chart will have (e.g. name, version, repository, values), even between versions and use that to locate the LCM charts. This is fine, as if we decide to remove name, version, repository or values this would be a breaking change way before it gets to LCM.
> > And what if we want/need to change the static definition at some point?

Then this is a breaking change that should be communicated in the documentation. However, that is why I suggest definitions that have a very low probability of changing. I recognise that this is not the best approach, but it is a compromise we could make. As I said we have a chicken-egg problem that needs to be resolved somehow, I just gave a suggestion on a possible approach..

> And if this specific portion of the manifest changes? Also, the versions of release manifest not staying compatible is exactly what we are basing this discussion on, no? I'm basing these two statements on below point:

Again, parsing a portion of the manifest as opposed to the full manifest significantly reduces the risk. Also, doing a lenient parse, where we do not break on unknown fields reduces the risk further. The only thing that could break us is the removal of, for example - "name", "version", "values", "repository". I do not see this happening, as this would merit a huge rework on how we handle helm charts, which in itself would be a serious breaking change on elemental level itself.

>I feel that if we introduce complete incompatibility between release manifests, neither currently proposed approach nor the one you are suggesting will help us do an upgrade.

Complete incompatibility would suggest drastically different release manifest. I am not talking about that, I am talking about how a minor change between manifests could break the manifest parse. 

For example -
1. Release `A` is the standard current release manifest.
2. Release `B` introduces a new optional filed called `foo`.

An upgrade for a Release `B` containing the filed `foo` will always fail if we parse it with the old LCM version, as that version be using an elemental library that is not aware of this new field. And since we always parse with [`decoder.KnownFields(true)`](https://github.com/SUSE/elemental/blob/main/pkg/manifest/api/shared.go#L81), this would cause an error, resulting in us never getting to the actual LCM upgrade.

Back to your statement, I wouldn't call the above example completely incompatible, for instance if we just had parsed the Helm portion of the manifest we would have never hit the above mentioned error, as we would have never parsed the full manifest.

> If release manifest versions are expected to introduce breaking change, LCM should be first updated to understand/parse the newer versions so that when release manifest breaking change is rolled out, things don't fall apart.

From my PoV, doing an intermittent upgrade of LCM, be it manual or through a different manifest would put unnecessary pressure to the user and would directly go against the "self-upgrading" methodology that we are going for. Also, as I said previously, it is my belief that we would not always want to back port such changes to older LCM versions. 

>I believe "LCM self-upgrade" means LCM updating things using one of the three kinds of reconcilers in the Pipeline. Custom logic to upgrade itself can't be called "self-upgrade".
And it's fine if we go down that route, just saying that it's not self-upgrade.

How is that different that what is introduced in this PR? We introduce a new reconciler (custom logic) to handle the upgrade. Regarding the custom logic, that is why [here](https://github.com/SUSE/elemental-lifecycle-manager/pull/97#discussion_r4006780980) I suggested to reuse the existing HelmReconciler, exactly so that we do not have custom upgrade logic. 

As far as what I think a "self-upgrade" is, I try to look at it from a user PoV.  As a user I would expect to go from release A to release B without first having to manually either upgrade LCM or go through an intermittent release.

I am happy to discuss this on a design discussion as well, but I am not convinced that we should be using the old LCM version to parse both the release manifest and upgrade configuration before we do the actual LCM upgrade.

internal/upgrade/reconcilers/lcm.go
ipetrov117: I agree, for specific charts we should always fail. This goes back to how we handle different charts failing, but that is a topic for a different discussion 😄

dharmit:
@ipetrov117 I'm still looking at other points in your response, but the first one is not making sense to me so replying to it first.

> > IIUC, this would cause a problem even if we move the upgrade logic to a different place in the flow.
> 
> Moving the upgrade logic before parsing the release manifest would fix this exact issue, as we would only parse the full release manifest once we have upgraded LCM itself first.

But how would we parse the new release manifest through the LCM controller that's already running on the cluster? I don't understand how you sugest to accomplish this.

1. When a new `Release` resource is detected by LCM controller, it will pull the container image mentioned in it.
2. After pulling the image, it will try to parse the release manifest stored at `/release-manifest.yaml`.

If this manifest schema changes in such a way that the running LCM controller can't parse it, how does it matter if we put the logic early in the flow or later?

ipetrov117:
@dharmit, I am sorry I was not able to convey what I mean in a more understandable manner, I'll try again.

> But how would we parse the new release manifest through the LCM controller that's already running on the cluster? I don't understand how you sugest to accomplish this.

What I suggest is **not** to parse the full manifest, but to parse just the Helm section of it and even better just the `name`, `version`, `values` and `repo` fields. We do a lenient parse, where we do not fail against any other fields we do not know. We trigger the LCM upgrade based on those values without doing the full parse of the manifest. This ensures two things:
1. We limit the collision to these four fields.
2. We do not do a full manifest parse before upgrading LCM first.

> If this manifest schema changes in such a way that the running LCM controller can't parse it, how does it matter if we put the logic early in the flow or later?

I am not saying that we will never hit a collision. What I am trying to do is minimise the change that such a collision will ever occur. From my PoV renaming or removing `name`, `version`, `values` and `repo` from the Helm setup would only ever happen if we completely rework our approach to Helm chart definitions. While not impossible, it is improbable.. Furthermore,  accepting this small risk, IMO, offers way better UX than the user having to do intermittent LCM upgrades on every LCM version bump.

Please take a look at the full comment and if you are still not convinced, we should talk about it on our design discussion call, so that we can come to a decision.

dharmit:
@ipetrov117 I see value in separating LCM upgrade from the current approach of using `upgrade.Pipeline`. I intend to have another top-level reconciler like: https://github.com/SUSE/elemental-lifecycle-manager/blob/dae79fc82a9c3d539c412d224e2141e8350fa580/internal/controller/release_controller.go#L71 which will only work with `HelmReconiler` with LCM charts in focus. Also, this reconciler will be registered in `main.go` before the `ReleaseReconciler` here https://github.com/SUSE/elemental-lifecycle-manager/blob/dae79fc82a9c3d539c412d224e2141e8350fa580/cmd/main.go#L204

Let me know if you have any concerns/suggestions. Mentioning this because I'd like to avoid a rework if a different/better approach is there in your mind.

ipetrov117:
> Let me know if you have any concerns/suggestions. Mentioning this because I'd like to avoid a rework if a different/better approach is there in your mind.

@dharmit, IMO reusing the existing HelmReconciler (as mentioned [here](https://github.com/SUSE/elemental-lifecycle-manager/pull/97#discussion_r4006780980)) would be a better fit than just wrapping the HelmReconciler in another reconciler. Let me know what you think.

> which will only work with HelmReconiler with LCM charts in focus. Also, this reconciler will be registered in main.go before the ReleaseReconciler here

I am not sure I get this, you want a totally different reconciler to do the actual LCM upgrade? Isn't that a bit too much? Why can't we just use the existing ReleaseReconciler and just have the LCM handling at the top before the manifest parse?

dharmit:
@ipetrov117 I meant to use the `HelmReconciler` that we currently have, albeit with an additional boolean field to indicate if it's supposed to be strict or not. And sure, I'll look at plugging this into the place where you've recommended. 👍🏽

